Issue #15, opened 2019-02-25
Finish line functionality


Zemanzo:
**Description**
Finish line at the end of the track that marbles will cross and mark them as finished.

**Minimum Acceptance Criteria**
Use the serialized data from a map file to determine the zone boundaries.
Timing for marbles will stop and mark them as finished when they enter this trigger.
Should have some kind of visual.

**Is blocked by the following issues**
#5 
#10

PR #48, opened 2019-03-07
Basic marble finishing & round automation

### LINKED ISSUE #16: Automated races
Races will be held at a regular interval, players will have a designated entering period, after which one can no longer join and the gate will open. The marbles will roll down for a maximum set amount of time, or until a certain period after the first marble has finished. After this, results are displayed and the cycle starts anew.

Zemanzo:
The marble finishing criteria are *incredibly* basic as of now, it checks all marbles 20 times per second whether their Y-value is below a certain threshold. Because of this, I will **refrain** from closing the related issue (#15).
Round automation has been reworked and every state change now gets sent over to the client. Although results are not yet displayed at the end, building this in will be trivial once the time is there so I will close #16 with this merge request.

~~Please don't look what happened with the intermediary commits, they are a mess...~~


public/scripts/chat/chat.js
```diff
@@ -218,14 +218,14 @@ function isAuthorized(data) {
 }
 
 /* Popout variant */
-if (window.user_data) {
+if (user_data) {
```
Qaomen: `user_data` doesn't follow naming conventions, and it's not clear where the variable comes from (ESlint gives a warning here).

public/scripts/client/game.js
```diff
@@ -2,97 +2,166 @@ import domReady from "../domReady";
...
 
 let audio = {
-	start: new Audio("resources/audio/start.mp3"),
-	end: new Audio("resources/audio/end.mp3")
-};
+		start: new Audio("resources/audio/start.mp3"),
+		end: new Audio("resources/audio/end.mp3")
+	},
+	marbles = [],
+	state = {};
```
Qaomen: I strongly recommend declaring `state`'s properties here beforehand. This gives an easy overview of what `state` contains at run-time.
Maybe consider making the property names more distinct to avoid confusion (reffering to vars such as `timerStart` and `roundTimerStart`, `state.roundTimer` as bool and `roundTimer` as function).

public/scripts/client/game.js
```diff
@@ -2,97 +2,166 @@ import domReady from "../domReady";
...
-		console.log(s, timeLeft);
-		let timerInterval = setInterval(function() {
-			if (timeLeft < 0) {
-				clearInterval(timerInterval);
-			} else {
-				timerElement.innerHTML = timeLeft;
-			}
-			timeLeft--;
-		}, 1000);
+let _timerInterval,
```
Qaomen: What is the underscore in the name for?

src/game.js
```diff
@@ -1,113 +1,175 @@
...
+module.exports = function(config, physics) {
+	return {
+		// Game logic
+		state: "started", // "waiting", "enter", "starting", "started"
+		setState(newState) {
+			this._socketManager.emit(newState, "state");
+			this.state = newState;
+			log.info("Current state: ".magenta, this.state);
+		},
+		_startDelay: 2825, // length in ms of audio
```
Qaomen: What are the underscores for?

src/game.js
```diff
@@ -1,113 +1,175 @@
 const log = require("./log");
 
-module.exports = function(config, physics, socketManager) {
-	// Game logic
-	let game = {
-		logic: {
-			state: "started" // "enter", "started"
+module.exports = function(config, physics) {
```
Qaomen, line 3: It was something we talked about elsewhere at a later date, but maybe we can use a constructor here and declare functions on the prototype to make this look a bit cleaner?

Qaomen (changes requested):
Some things could be changed, but it's probably best to discuss them and see how necessary they are.

public/scripts/chat/chat.js
Zemanzo: Using `window.user_data` was an attempt to make it clear its a global variable that gets injected in the global scope. It's a bit tricky because it is a variable that comes from the server and has to be sent to the client. (See the `chat.mustache` template). I do agree this should be done better, but at this point I am unsure how. Naming convention comes from the API which calls it this, but I suppose we can change that.

public/scripts/client/game.js
Zemanzo: Underscores are for marking variables as private. It doesn't change anything technically, but for lack of anything better, this is a common convention to follow.

src/game.js
Zemanzo: Private variable (see other comment)

public/scripts/client/game.js
Zemanzo: Agreed, I'll get on this

public/scripts/chat/chat.js
Qaomen: I'd prefer `window.user_data` in that case so it's indeed clear where it comes from (and it keeps the linter from popping up). Perhaps it's for the best to leave the name itself as is then, to show the connection to the API.

Zemanzo: The only thing is that it seemed to break with `window.user_data`, that's why I reverted it again... ;-;

Qaomen (approved):

PR #49, opened 2019-03-11
Maximum marble count & basic client HUD notification system

Zemanzo:
Merge #48 before merging this.
Closes #18

PR #52, opened 2019-03-12
Graceful shutdown

### LINKED ISSUE #18: Maximum marble count
Have a maximum amount of marbles to avoid crashes & Warn players trying to join after the maximum has been hit.

Zemanzo:
This does not consider currently in-progress races yet, it's merely technical. A basic notification is sent to the client, but this can be improved.

Contains changes from #49 
Closes #18


Qaomen (changes requested):
The maximum marble count check doesn't apply to bot marbles, but it probably should. Maybe make it a maxPlayerCount instead.

Qaomen (approved):

PR #55, opened 2019-03-14
Feature/editor refactor

Qaomen:
Major refactor with too many things changed to mention.
Some things to note however: Center model origin and the terrain collider are currently disabled. They need reworking in order for them to function properly, which will be done at a later time.


Zemanzo (approved):
Awesome work! There's still a few minor issues but I'd rather solve those in other PRs.

PR #65, opened 2019-03-19
Reworked config file with user overriding properties

Qaomen:
Slightly reworked server config file: It's now possible to override specific properties via the userConfig. This should minimize manual config.js updates.


Zemanzo (changes requested):
I'd like a big ol' warning that users should never change the config.js directly commented inside the file

Zemanzo (approved):

PR #105, opened 2019-03-27
Patch/editor fixes

Qaomen:
Fixes non-existing transform properties and changed prefab entity type to lower-case.


Zemanzo (approved):
Have not actually confirmed these working, but I'm pretty sure it's fine. At the very least, the editor doesn't seem broken so that's good.

Issue #70, opened 2019-03-20
Refactor: Physics


Qaomen:
**Description**
To clean up the code and in preparation for in-editor level testing, the current physics code should be refactored and moved to `src/physics`, so it's accessible for both the editor and server.

Note: This could become part of a much larger refactor, depending on how the game logic will be handled in the future.

**Minimum Acceptance Criteria**
- There should be a `src/physics` folder containing the physics-related scripts.
- Any code specific to the server implementation (e.g. the `getFinishedMarbles` logic) should be moved into their relevant script(s).
- The physics code should follow the [Style Guide](https://github.com/Zemanzo/web-marbles/wiki/Style-guide).

**Is blocked by the following issues**
Although not blocked, #58 may become related to this refactor.

PR #123, opened 2019-04-02
Level loading

### LINKED ISSUE #6: Level loading
A single endpoint that takes in a file path to the map file and parses it to both a physics environment for the server, and a visual environment to be sent to the client.
- [x] Server physics
- [x] Server gameplay parameters
- [x] Client models
- [x] Client water / sun parameters
- [x] Remove old terrain

#5

### LINKED ISSUE #109: World hierarchy doesn't scroll
**Expected behavior**
When there are too many world objects in the hierarchy to fit on the screen, a scroll bar should appear that allows the user to scroll up/down.

**Actual behavior**
No scroll bar appears, and navigating up/down isn't possible.

**Reproduction steps**
1. Go to the editor's World tab.
2. Add a bunch of prefabs until the hierarchy list becomes too long.

**Additional context**
Functionality should be similar to the prefabs list, or the entity list inside the prefab itself.

Zemanzo:
Resolves #6 
Resolves #109


src/client/client/render.js
```diff
@@ -3,14 +3,16 @@ import "three/examples/js/objects/Water";
 import "three/examples/js/objects/Sky";
 import "three/examples/js/loaders/LoaderSupport";
 import "three/examples/js/loaders/OBJLoader";
+import "three/examples/js/loaders/GLTFLoader";
 import "three/examples/js/nodes/THREE.Nodes";
 import * as Stats from "stats-js";
+import * as pako from "../../../node_modules/pako/dist/pako";
```
Qaomen: Why not `import * as pako from "pako"`? Regardless, this might be something to put into the style guide.

src/client/client/render.js
```diff
@@ -394,10 +283,98 @@ let clearMarbleMeshes = function() {
...
+								}
+							}
+
+							resolve(group);
+						});
+					}
+				}).then(() => {
+					Promise.all(Object.values(prefabPromises)).then(() => {
+						for (let object of Object.values(mapData.worldObjects)) {
+							console.log(prefabPromises, object.prefab);
```
Qaomen: Maybe best to remove the logging on success in this function, even if it's due for refactor.

src/server/maps/manager.js
```diff
@@ -0,0 +1,63 @@
...
+
+// Scan for maps
+function retrieveMaps() {
+	return new Promise((resolve, reject) => {
+		fs.readdir(config.maps.folderPath,
+			undefined,
+			function(error, files) {
+				if (files && Array.isArray(files)) {
+					// Only read files that have the correct extension
+					let mapFiles = files.filter(file => file.endsWith(".mmp") || file.endsWith(".mmb"));
```
Qaomen: Only map files ending with `.mmb` are supported here.

src/server/maps/manager.js
```diff
@@ -0,0 +1,63 @@
...
+					if (mapFiles.length > 0) {
+						return resolve(mapFiles);
+					}
+				}
+
+				log.error("No files found");
+				reject("No files found");
```
Qaomen, line 21: Rejections aren't handled and crash the server. We could consider this part of a future issue on loading levels with level rotation, though.

src/server/physics/map-builder.js
```diff
@@ -0,0 +1,96 @@
+module.exports = function(Ammo, world, map) {
+	let _startAreas = [],
+		_endAreas = [];
```
Qaomen: Start/end areas should be part of the world (like starting gates).

src/server/physics/map-builder.js
```diff
@@ -0,0 +1,96 @@
...
+		for (let worldObjectUuid in map.worldObjects) {
+			let worldEntity = map.worldObjects[worldObjectUuid];
+			let prefab = map.prefabs[map.worldObjects[worldObjectUuid].prefab];
+
+			for (let prefabEntityUuid in prefab.entities) {
+				let prefabEntity = prefab.entities[prefabEntityUuid];
+
+				let worldEntityTransform, prefabEntityTransform;
+
+				//console.log(prefabEntity.type);
```
Qaomen: Can be removed.

src/server/physics/map-builder.js
```diff
@@ -0,0 +1,96 @@
+module.exports = function(Ammo, world, map) {
+	let _startAreas = [],
+		_endAreas = [];
+
+	map.currentMapData.then((map) => {
+		function transformFromEntity(entity) {
```
Qaomen: I think it would be cleaner to make this a private function instead of defining it here.

src/server/physics/marbles.js
```diff
@@ -1,10 +1,39 @@
-module.exports = function(Ammo, world) {
+module.exports = function(Ammo, world, mapBuilder) {
+	let _marblesTransformAux = new Ammo.btTransform(),
+		_pos = undefined,
+		_rot = undefined,
```
Qaomen: Still need to update the conventions for it, but best use `null` instead of `undefined` here, so it's clear it's a used variable.
Edit: We actually still use it in several places, so instead of going over all of them I'd say fix when you see them is fine too?

src/server/physics/marbles.js
```diff
@@ -1,10 +1,39 @@
...
+
 	return {
 		list: [],
-		_marblesTransformAux: new Ammo.btTransform(),
-		_pos: undefined,
-		_rot: undefined,
-		_id: 0,
 
 		createMarble(meta) {
 			// Create physics body
```
Qaomen, line 39: Digression: We should have the size randomness defined in `meta`, not in this function.

src/server/physics/world.js
```diff
@@ -1,149 +1,230 @@
...
-	// Physics letiables
-	let collisionConfiguration,
-		dispatcher,
-		broadphase,
-		solver,
+	let _collisionConfiguration,
+		_dispatcher,
+		_broadphase,
+		_solver,
 		physicsWorld;
```
Qaomen, line 6: ~~Should be prefixed with an underscore.~~ Nevermind, just realised it's a sort of forward declaration. It might be a good idea to add a comment for that.

src/server/physics/world.js
```diff
@@ -1,149 +1,230 @@
...
-		let flipQuadEdges = false;
+	_startUpdateInterval();
+
+	return {
+		map: undefined,
+		physicsWorld,
+		updateInterval: undefined,
+		gates: [],
+
+		toggleGates(override) {
```
Qaomen: I think this function could be a bit better named/used. Maybe `setGateStates` where the parameter could be `"open"` or `"close"` and is explicitly handled in both of these cases.

public/styles/editor.css
```diff
@@ -442,14 +416,18 @@ body {
...
+	display: flex;
+	flex-direction: column;
 }
 
 	#worldHierarchy {
 		background: #111;
 		border: 3px solid #262626;
 		padding: 2px;
-
+		overflow-y: auto;
```
Qaomen: Maybe if would be better to have the "Add Prefab" always visible, either on top or bottom of the hierarchy list.

Qaomen (changes requested):
Don't pay too much mind to all the comments - lots of good stuff in this PR! :D 
Most things are minor, or up to discussion, or things that we may better take care of in other Issues.

src/client/client/render.js
Zemanzo: Remainder of a small time period where I thought there was a difference between browser supported pako versions and server-side pako versions... I'll change this back to the usual.

src/server/physics/marbles.js
Zemanzo: I'll leave that for #70 ;)

src/server/maps/manager.js
Zemanzo: Yeah we'll bother with this later, once we figured out how exactly we want our exceptions

Qaomen (approved):

PR #128, opened 2019-04-03
Marbles no longer stay frozen when the race starts

### LINKED ISSUE #120: Marbles don't move when the race starts
**Expected behavior**
Once the starting gates disappear and the race starts, the marbles will start rolling down the track.

**Actual behavior**
The marbles don't move, and a Nightbot marble is dropped to awaken the sleeping marbles. In the case of multiple starting areas, some marbles may start frozen for the entire race.

**Reproduction steps**
Use eyes.

**Additional context**
We already know why it happens and how to fix it but I'm making it an issue anyway so everything isn't all blobbed together in an amalgamation of relative awesome.

Qaomen:
Closes #120

### DIFF
--- a/src/server/game.js
+++ b/src/server/game.js
@@ -177,9 +177,9 @@ let game = function() {
 					this.startTime = Date.now();
 
 					physics.world.setAllGatesState("open");
-
-					// Add bot marble to ensure physics not freezing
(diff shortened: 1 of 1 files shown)


Zemanzo (approved):
RIP OG Nightbot :'(

PR #129, opened 2019-04-03
Editor transform fixes and changes

Qaomen:
No Issues to reference since there are none! But there sure were some things worth changing:
- Fixed object position/scale/shape properties being set to NaN, caused by invalid input fields.
- Fixed a bug where box shape properties were reset when changing collider function to Starting Area.
- Changed the Euler rotation order from `XYZ` to `YXZ`, which means the following:
    - Possibly a bit more intuitive for track prefabs: First the direction (Y), and then the slope pitch/roll (XZ) rotations are applied.
    - No gimbal lock when Y is 90°, which feeds into the reasoning above. (It's on X=90° now, which is much less likely.)
    - Level files and loading is unaffected since they're stored in quaternions. The Euler angle values may change when loading it in the editor but the rotations remain the same.

### DIFF
--- a/src/client/editor/inspector.js
+++ b/src/client/editor/inspector.js
@@ -136,25 +136,33 @@ let inspector = function() {
 
 			// Translation
 			transformFunctions.translate = function(axis, value) {
+				if(isNaN(value)) value = 0;
 				let position = inspector.selected.getPosition();
-				position[axis] = parseFloat(value);
+				position[axis] = value;
 				inspector.selected.setPosition(position);
 			};
 
 			// Scale
 			transformFunctions.scale = function(axis, value) {
+				if(isNaN(value)) value = 1;
 				let scale = inspector.selected.getScale();
-				scale[axis] = parseFloat(value);
+				scale[axis] = value;
 				inspector.selected.setScale(scale);
 			};
 
 			// Rotate
 			transformFunctions.rotate = function() {
 				let rotation = inspector.elements.transform.input.rotate;
+				let x = rotation.x.valueAsNumber;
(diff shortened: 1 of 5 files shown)


src/client/editor/inspector.js
```diff
@@ -134,26 +136,34 @@ let inspector = function() {
...
-				scale[axis] = parseFloat(value);
+				scale[axis] = value;
 				inspector.selected.setScale(scale);
 			};
 
 			// Rotate
 			transformFunctions.rotate = function() {
 				let rotation = inspector.elements.transform.input.rotate;
+				let x = rotation.x.valueAsNumber;
+				if(isNaN(x)) x = 0;
```
Zemanzo: Since (from what I've read) `.valueAsNumber` always returns a valid number OR NaN otherwise, wouldn't it be nicer to simply write `let x = rotation.x.valueAsNumber || 0`?

Zemanzo (changes requested):

src/client/editor/inspector.js
```diff
@@ -194,22 +204,30 @@ let inspector = function() {
 
 			// Width
 			shapeChangeFunctions.width = function() {
-				inspector.selected.setWidth(parseFloat(inspector.elements.shapeProperties.input.width.value));
+				let value = inspector.elements.shapeProperties.input.width.valueAsNumber;
+				if(isNaN(value)) value = 1;
```
Zemanzo: You can't do that same trick here though (because if `value === 0`, it will be still be a falsey value)
Additionally, should we verify these values to make sure they're not 0 or negative?

Zemanzo (approved):

PR #143, opened 2019-04-24
Mesh colliders and double-file publishing

### LINKED ISSUE #23: Importing physics colliders
**Description**
It should be possible to load models in the editor and use them as colliders (concave and convex).

- Support for convex colliders.
- Support for concave colliders.
- Models UI should be expanded to indicate support for these features per-model.

**Stretch goals**
- Support chunked terrain / height maps.
- Models that are used purely as colliders could have an optimized format in the level file.

#70 #77

Qaomen:
Adds the ability to select colliders based on loaded models in the editor. Also includes splitting exports into a client-side and server-side file, and optimising the contents for each.

**Referencing issues**
Closes #23


src/client/editor/models.js
```diff
@@ -3,12 +3,17 @@ import "three/examples/js/loaders/GLTFLoader";
...
 import { scene } from "./render";
+import "three/examples/js/QuickHull";
+import "three/examples/js/geometries/ConvexGeometry";
 
 // model object
-function Model(name, sceneObject) {
+function Model(name, sceneObject, project) {
 	this.name = name;
 	this.sceneObject = sceneObject;
+	this.project = project;
```
Zemanzo: This being called `project` is pretty confusing, I though it was a reference to a global `Project` object or variable at first. Maybe call it `projectData`?

Zemanzo: And as you mentioned, rename `project.data` to `projectData.file`

src/client/editor/prefabs.js
```diff
@@ -506,17 +514,70 @@ PrefabCollider.prototype.setRadius = function(radius) {
 	}
 };
 
+PrefabCollider.prototype.setModel = function(modelName) {
+	if(this.colliderData.shape !== "mesh") return;
+
+	// Remove from references
+	if(this.colliderData.model) {
+		delete modelsTab.models[this.colliderData.model].prefabEntities[this.uuid];
+		this.sceneObject.geometry = new THREE.Geometry();
```
Zemanzo, line 523: Any specific reason for using THREE.Geometry over `THREE.BufferGeometry`? As per the documentation:
>Geometry is a user-friendly alternative to BufferGeometry. Geometries store attributes (vertex positions, faces, colors, etc.) using objects like Vector3 or Color that are easier to read and edit, but less efficient than typed arrays.
>
> Prefer BufferGeometry for large or serious projects.

https://threejs.org/docs/#api/en/core/Geometry

src/client/editor/prefabs.js
```diff
@@ -506,17 +514,70 @@ PrefabCollider.prototype.setRadius = function(radius) {
...
+	if(isConvex) {
+		this.sceneObject.geometry = modelsTab.models[this.colliderData.model].getConvexHull();
+	} else {
+		this.sceneObject.geometry = modelsTab.models[this.colliderData.model].getConcaveGeometry();
+	}
+	if(!this.sceneObject.geometry) {
+		editorLog(`Failed to switch collider to ${isConvex ? "convex" : "concave"}.`, "error");
+		// Reset collider data so it stays valid
+		delete modelsTab.models[this.colliderData.model].prefabEntities[this.uuid];
+		this.sceneObject.geometry = new THREE.Geometry();
```
Zemanzo, line 560: No `BufferGeometry`?

Zemanzo (changes requested):
Overall, it's great stuff once again. Keep it up!

src/client/editor/prefabs.js
Qaomen: In this case it's empty placeholder geometry, since setting it to null causes rendering to fail. It's replaced by `THREE.BufferGeometry` in other cases. (Except for mesh colliders actually, so that'll need a change!)

Zemanzo (approved):

Issue #81, opened 2019-03-22
Ability to toggle between cameras


Zemanzo:
**Minimum Acceptance Criteria**
- [x] A UI element that allows selection of the preferred camera mode.
- [x] Clicking the corresponding UI element switches the viewport to the related camera.

**Additional notes**
Support for hotkeys would be a plus (F switches to the free camera, G to the Global, T to the tracking?)

**Is blocked by the following issues**
Need at least two different cameras, so #31 and #32 most likely

PR #149, opened 2019-04-30
Render refactor & dependency updates

### LINKED ISSUE #3: Freecam (pointerlockcontrols) movement angle does not correspond with looking angle
**Expected behavior**
When using the freely controlled camera, moving the camera forward by holding the `W` key should move the camera in the direction it is facing at that moment.

**Actual behavior**
When looking straight up or down, moving forward (hold W) does not move the camera straight up or down like it should, but instead moves it slightly forward still.

**Reproduction steps**
1. Go to the client or editor
2. Angle the camera fully downwards or upwards
3. Hold W

### LINKED ISSUE #68: Game viewport does not resize with browser viewport
**Expected behavior**
When resizing the browser viewport, the game viewport should resize accordingly (shrink when smaller, grow when larger)

**Actual behavior**
The game viewport remains at the size it initialized on, causing overflowing or space not being utilized.

**Reproduction steps**
1. Go to /client
2. Resize the browser viewport

### LINKED ISSUE #69: Refactor: Renderer
**Description**
Currently, both the client and editor have their own renderer script in their own respective folders. Portions of it are identical to one another while other parts are client- or editor-specific. The base renderer code should be split and brought under a new `render` folder that both the client and editor will use.

- There should be a `src/client/render` folder containing the renderer-related script(s).
- Any code specific to the client or editor (e.g. the grid in the editor) should be moved into their relevant scripts in `src/client/client` or `src/client/editor`.
- The renderer code should follow the [Style Guide](https://github.com/Zemanzo/web-marbles/wiki/Style-guide).
- Existing WebGL warnings should be resolved, if possible.

Although not blocked or directly related, #68 could be taken into account during the refactor.

Zemanzo:
With this PR, the client and the editor share a rendering script, so that there no longer is duplicate code for rendering. It should also get rid of all WebGL warnings.

The gist of the new system is that there is a main scene that gets rendered which contains only the camera and marbles, and a child scene for every map (although usually that's only a single map). A map contains all aesthetics, including the skybox and water plane.

Closes #68 
Closes #69 
Also magically fixes #3 (because it seemed fine?) although I have no idea why.


src/client/render/render-core.js
```diff
@@ -0,0 +1,350 @@
+import * as THREE from "three";
+import "three/examples/js/objects/Water";
+import "three/examples/js/objects/Sky";
+import "three/examples/js/loaders/LoaderSupport";
+import "three/examples/js/loaders/GLTFLoader";
```
Qaomen, line 5: `LoaderSupport` and `GLTFLoader` are unused.

src/client/render/render-core.js
```diff
@@ -0,0 +1,350 @@
...
+const _isWebGLAvailable = function() {
+	try {
+		const canvas = document.createElement("canvas");
+		return !!(window.WebGLRenderingContext && (canvas.getContext("webgl") || canvas.getContext("experimental-webgl")));
+	} catch (error) {
+		return false;
+	}
+};
+
+let _mainScene = null,
```
Qaomen: The renderer should probably use the module pattern, especially since there are some classes in here as well that aren't part of the module. Could improve readability overall too.

src/client/render/render-core.js
```diff
@@ -0,0 +1,350 @@
...
+	_renderer = null,
+	_marbles = {
+		meshes: []
```
Qaomen: Why not make `_marbles` the array?

src/client/render/render-core.js
```diff
@@ -0,0 +1,350 @@
...
+	},
+	animationUpdateFunctions = [],
+	_viewport = null, // DOM viewport element
+	_axesHelper = null,
```
Qaomen: Minor thing, but I don't think the client needs it.

src/client/render/render-core.js
```diff
@@ -0,0 +1,350 @@
...
+	_stats = null,
+	_controls = {
```
Qaomen: Since `fallback` is unusued / the same as `active`, it may be better to only use `_controls` instead.

src/client/render/render-core.js
```diff
@@ -0,0 +1,350 @@
...
+	this.scene.add(this.water.waterObject);
+	this.sky.water = this.water;
+
+	// Update sky
+	this.sky.update();
+}
+
+MarbleMap.prototype.addToWorld = function() {
+	_mainScene.add(this.scene);
+	console.log(_mainScene);
```
Qaomen: Leftover logging.

src/client/render/render-core.js
```diff
@@ -0,0 +1,350 @@
...
+		return false;
+	}
+};
+
+let _mainScene = null,
+	_renderer = null,
+	_marbles = {
+		meshes: []
+	},
+	animationUpdateFunctions = [],
```
Qaomen: Clogging this comment with some bigger structural stuff:
I think we can make this work without this array of functions. Right now it's used for water and marble updates (client only). The marble updates is actually part of a bigger problem I think. What's currently in place works but at the same time feels like a band-aid solution, and once we work on in-editor testing it'll probably get more complex. We probably need to look into this.

src/client/render/render-core.js
```diff
@@ -0,0 +1,350 @@
...
+}
+
+function _onCanvasResize() {
+	_renderer.setSize(_viewport.clientWidth, _viewport.clientHeight);
+
+	_controls.active.camera.aspect = _viewport.clientWidth / _viewport.clientHeight;
+	_controls.active.camera.updateProjectionMatrix();
+}
+
+function MarbleMap(data) { // "Map" is taken
```
Qaomen: `MarbleMap` could have an update function that includes all map-related stuff (which is only the water atm). I'm not sure what the intended use of `addToWorld` and `removeFromWorld` is yet, but since the renderer would update stuff and there would probably be only one map 'active' at a time, these things could tie in together nicely maybe?

src/client/render/render-core.js
```diff
@@ -0,0 +1,350 @@
...
+			fog
+		}
+	);
+
+	this.waterObject.rotation.x = -Math.PI / 2;
+	this.waterObject.position.y = waterLevel;
+	this.waterObject.material.uniforms.size.value = 8;
+
+	animationUpdateFunctions.push(() => {
+		this.update.call(this);
```
Qaomen: As mentioned before, having a `MarbleMap.update` to do this would probably look better (and keeps the class a bit more standalone too).

src/client/render/render-core.js
```diff
@@ -0,0 +1,350 @@
...
+	uniforms.mieCoefficient.value = 0.005;
+	uniforms.mieDirectionalG.value = 0.8;
+
+	this.parameters = parameters;
+	this.parameters.distance 	= !isNaN(parameters.distance)	 ? parameters.distance	  : 4000;
+	this.parameters.inclination = !isNaN(parameters.inclination) ? parameters.inclination : .25;
+	this.parameters.azimuth 	= !isNaN(parameters.azimuth)	 ? parameters.azimuth	  : .205;
+}
+
+Sky.prototype.update = function() {
```
Qaomen: For a second I thought this was called every frame since it's called `update`. Maybe change the name or add a comment? I might be a bit nitpicky here though.

src/client/render/render-core.js
```diff
@@ -0,0 +1,350 @@
...
+			_mainScene.remove(marble.mesh.children[i]);
+		}
+		_mainScene.remove(marble.nameSprite);
+		_mainScene.remove(marble.mesh);
+	}
+
+	_marbles.meshes = [];
+};
+
+const updateMarbleMeshes = _marbles.update = function(newPositions, newRotations, delta) {
```
Qaomen: Defining `_marbles.update` and `_marbles.removeAll` doesn't seem to be necessary.

Qaomen: Come to think of it this might also be better suited elsewhere since it's game-related and not rendering. But that's just a thought for later (and related to in-editor testing and the networking thing).

src/client/client/render.js
```diff
@@ -1,377 +1,119 @@
...
-
-	// Update controls
-	if ( controls.enabled === true ) {
-		controls.update();
+});
+
+/**
+ * Parses the map data and returns a Promise that resolves once it is fully done loading
+ */
+renderCore.MarbleMap.prototype.parseData = function() {
```
Qaomen: I really recommend keeping relevant code together. Modifying a class outside its source file is pretty confusing.

src/client/render/render-core.js
```diff
@@ -0,0 +1,350 @@
...
+
+function _onCanvasResize() {
+	_renderer.setSize(_viewport.clientWidth, _viewport.clientHeight);
+
+	_controls.active.camera.aspect = _viewport.clientWidth / _viewport.clientHeight;
+	_controls.active.camera.updateProjectionMatrix();
+}
+
+function MarbleMap(data) { // "Map" is taken
+	this.data = data;
```
Qaomen: Will `this.data` be used for anything? Since it contains defaults (or is used for loading a level), I don't think we'll need it in the end. `MarbleMap.parseData` could take the level data as an argument and discard it after, saves a bit of RAM.

src/client/client/render.js
```diff
@@ -1,377 +1,119 @@
...
+	let modelPromises = {};
+	for (let modelName in this.data.models) {
+		modelPromises[modelName] = new Promise((resolve, reject) => {
+			try {
+				_GLTFLoader.parse(this.data.models[modelName].file, null,
+					function(model) {
+						resolve(model.scene);
+					}, function(error) {
+						console.warn(`Unable to load model (${modelName})`, error);
+						reject("error");
```
Qaomen: These errors aren't caught, causing the rest of the loading to fail. For reference you could check the loading in `project.js`. Maybe replace unloaded models with the default model like in the editor?

src/client/client/render.js
```diff
@@ -1,377 +1,119 @@
...
-	let spriteMaterial = new THREE.SpriteMaterial({map: texture});
-	let sprite = new THREE.Sprite( spriteMaterial );
-	sprite.scale.set(4, 1, 1.0);
-	return sprite;
-}
+	// Load prefabs
+	let prefabPromises = {};
+	return Promise.all(Object.values(modelPromises)).then(() => {
+		for (let prefabUuid in this.data.prefabs) {
+			prefabPromises[prefabUuid] = new Promise((resolve) => {
```
Qaomen: I don't think these `Promise`s are strictly necessary since it's all synchronous anyway? But could be wrong since they're confusing lil things.

src/client/editor/render.js
```diff
@@ -1,156 +1,24 @@
...
-//
-
-uniforms = {
-	time: { value: 1.0 }
-};
-let clock = new THREE.Clock();
+map.addToWorld();
 
 // Default model
 let defaultModel;
```
Qaomen: Like mentioned in `client/client/render.js`, maybe have the default model as a global fallback model in render core?

src/client/editor/render.js
```diff
@@ -1,156 +1,24 @@
...
-			x: -2.3,
-			y: 12,
-			z: 19.7
-		},
-		defaultRotation: {
-			x: -.3,
-			y: 0,
-			z: 0
+const GLTFLoader = new THREE.GLTFLoader(),
+	map = new renderCore.MarbleMap({
```
Qaomen: Not sure how I feel about this. Maybe `renderCore` could have one active map (that could be switched out in case of client stuff) instead of having one defined loosely like this? Keeps it more global and contained, I dunno.

Qaomen (changes requested):
All the comments and such seem like much, but this is still improvement! It's got a lot of promising stuff imo, but overall it also seems plagued by loose ends and designs that aren't fully fleshed out yet.
With some explanations and/or discussions (and writing the TDD part on it?) things would be clarified on what the bigger picture of it all is.

(Oh, and the WebGL warnings are gone wooh!)

src/client/render/render-core.js
Zemanzo: I'm in the dark on what you mean with this. Could you please expand on it? How would you envision this?

Qaomen: The goal would be to have the functions that are called in the render loop clearly defined, so it's easier to see what happens during a render update. A simplified example would be this:
```js
function animate() {
  marbleManager?.updateMarbles(); // What this does and where the function comes from still unclear
  _controls.update(); // Might come from a Camera class later if we have tracking cams?
  activeLevel.update(); // Updates water and such
  _renderer.render();
}
```
Issue right now though, is that updating marbles is client-only (and depends on networking) which is done by adding it to the `animationUpdateFunctions` array somewhere. Once we have in-editor testing, those marbles need to be updated the same way but through physics.

We'd probably benefit from managing those marbles at the same level as renderCore (so it can call a function like above instead), but for that it needs some changes and planning ahead. How the marbles are updated internally (network or physics) would be handled in the manager and client/editor script(s).

src/client/client/render.js
Zemanzo: I think they're necessary, the docs do make it sound as if GLTFLoader.parse is rather asynchronous https://threejs.org/docs/#examples/loaders/GLTFLoader.parse

Qaomen: I meant for the code after the loading, sorry if that wasn't clear. It's good it's wrapped in the `.then`, but I don't think there's a need to create new `Promise`s for cloning the loaded scene and setting the transforms.

Zemanzo: Ah yes, this was actually due to me first having the idea to not actually wait for all models to load before starting to parse the prefabs, but only those required for those prefabs. In this situation, the asynchronous nature of the models would propagate to the prefabs and world objects.

Should I try to implement it that way, save it for later, or just get rid of the promises here and not worry about the minimal performance gain?

Qaomen: I like the idea (and seeing the map gradually load in like that would look cool too), but it's probably not worth doing right now. I'd remove the additional promises since they won't be needed.

src/client/render/render-core.js
Zemanzo: My intention here was to (later) add a way to switch between controls, although it should probably have been part of #81

Qaomen (approved):

PR #150, opened 2019-05-03
Discord optional when debugging

### LINKED ISSUE #125: Make Discord optional
**Description**
Currently, without a valid Discord configuration, the project cannot be run. This makes it hard for outside contributors to work on the client or editor.

The server needs to be able to run without having access to the Discord API.
The client, chat and editor still need to load. The client and chat do not have to be functional however.

Zemanzo:
This adds a configuration option where Discord integration can be disabled. **This should only be used during development as it is not secure.**

Closes #125


Qaomen (approved):
Quite the useful lil thing! :D

PR #163, opened 2019-05-21
Client-side starting gate visualisation

### LINKED ISSUE #14: Starting gate functionality
**Description**
Collider that moves away to let the marbles onto the track when the game starts. Should have a visual.

- [x] Use the serialized data from a map file to determine where the starting gate is located.
- [x] When the game starts, the gate should disappear. (Brownie points to make it fade slowly maybe?)
- [x] When marbles can be entered again, the gate should reappear.

#5
#10

Qaomen:
Basic disappearing/reappearing starting gate for the client.

**Referencing issues**
Closes #14


src/client/client/client.js
```diff
@@ -2,10 +2,32 @@ import domReady from "../dom-ready";
...
+	fetch(`/resources/maps/${levelName}.mmc`)
+		.then((response) => {
+			// Return as a buffer, since .text() tries to convert to UTF-8 which is undesirable for compressed data
+			return response.arrayBuffer();
+		})
+		.then((buffer) => {
+			let levelData = levelManager.load(buffer);
+			renderCore.activeLevel.loadLevel(levelData)
+				.then( () => {
+					if(game.getCurrentGameState() === "started") {
```
Zemanzo, line 28: I think it would be better to put this in `/game.js`, that's where all the other gameplay elements get initialized as well.

Zemanzo: Actually, I just noticed you already did that and why it doesn't work. Should we think of a better solution though? I'm not a fan of this, but on the other hand I can live with it and don't really feel like delaying other things.

Zemanzo (commented):
Just random brainthoughts, it's good!

src/client/client/client.js
Qaomen: Yeah, it's currently part of a bigger problem described in #164. Once async code and inter-dependencies are resolved, this should have a proper place. It's possible the `client.js` might be removed, since it doesn't have much going for it at the moment.

Zemanzo (approved):

PR #167, opened 2019-05-23
Custom materials

### LINKED ISSUE #152: Custom materials
**Description**
Since the GLTF specification does not support all types of materials, a custom way to create materials is desired.

- [x] Custom materials can be created in the editor by writing a small script and parsing it.
- [x] Said materials should be exported along with the rest of the editor project.
- [x] It should be possible to use custom textures as well.
- [x] Custom materials can be applied to any imported model.
- [x] Custom materials should load and apply automatically when loading a level.

### LINKED ISSUE #154: Improve water reflection
**Description**
Currently, the water (or ocean) dynamically reflects everything in the scene, including the marbles. This is devastating for performance, and thus must be altered.

- [x] Water should never reflect name tags.
- [x] Have a flag to disable the reflection of marbles.
- [x] Have a flag to disable the reflection of the level.

**Related issues**
Toggling it on or off is discussed in #155

Zemanzo:
Oh baby, it's a big one.

**Editor**
Updated to v0.3.0

This adds two new tabs in the editor:
- Textures, which allow you to upload images as textures.
- Materials, where you can create a customizable material using the uploaded textures.
It also adds extra functionality to the models tab, where you can select a custom material for each of the child meshes a model has.

Under the hood, editor HTML elements have been generalized. A new `item` class can be used for a generic expandable element as seen before with prefabs.

For the models tab, `removeModel` has been removed in favor of `Model.delete()`.

**Client**
Loads custom materials if they exist

**Other**
Water only reflects the skybox.
Old map files and models have been removed.

**Issues**
Closes #152


src/client/editor/textures.js
```diff
@@ -0,0 +1,144 @@
...
+import { modelsTab } from "./models";
+import { projectTab } from "./project";
+import { editorLog } from "./log";
+import { generateTinyUUID } from "../generate-tiny-uuid";
+
+// texture object
+function Texture(uuid, name, texture, projectData) {
+	this.uuid = uuid;
+	this.name = name;
+	if (!projectData.name) projectData.name = name;
```
Qaomen: Project data should always be valid/complete, so it shouldn't be necessary to add a property elsewhere to make it valid. I recommend passing the name to `Level.addTexture` instead.

src/client/editor/textures.js
```diff
@@ -0,0 +1,144 @@
...
+	this.map = new THREE.TextureLoader().load(texture);
```
Qaomen: Wouldn't it make more sense to call it `this.texture`?

src/client/editor/materials.js
```diff
@@ -1,39 +1,256 @@
...
+	this.optionElements = [];
+
+	// Set a default material
+	this.compiledMaterial = new DefaultMaterial();
+
+	// Add events
+	this.elements.itemName.addEventListener("input", function() { self.onNameChange(this.value); }, false);
+	this.elements.itemName.addEventListener("change", function() { self.onNameChange(this.value); }, false);
+	this.element.getElementsByClassName("collapse")[0].addEventListener("click", function() { self.toggleCollapse(); }, false);
+	this.element.getElementsByClassName("delete")[0].addEventListener("click", function() { self.delete(); }, false);
```
Qaomen: No confirmation prompt.

src/client/editor/textures.js
```diff
@@ -0,0 +1,144 @@
...
+		if (confirm(`Are you sure you want to delete texture ${name}?`)) {
+			texturesTab.removeTexture(uuid);
+		}
+	}, false);
+
+	// Add to DOM
+	this.element = document.getElementById("textures").appendChild(this.element);
+
+	// Add texture option to each material selection
+	for (let uuid in materialsTab.materials) {
```
Qaomen, line 49: This works, but feels like it could be cleaner. I'll leave it to you whether it's something worth changing, but some thoughts:
Is there a way to have one set of selections that all 5 material properties (per material) point to, instead of filling 5 lists for each? If so, `materialsTab` could just maintain one list, instead of managing it at texture creation/deletion and material creation. The code's a bit spread out at this point.
If a material is removed, `Texture.optionElements` isn't updated. It doesn't give warnings when you delete said textures, but it does look a bit messy this way.

src/client/editor/textures.js
```diff
@@ -0,0 +1,144 @@
...
+						let project = projectTab.activeProject.addTexture(uuid, file.reader.result);
+						texturesTab.addTexture(uuid, file.name, file.reader.result, project);
+					};
+
+					file.reader.onerror = function() {
+						editorLog(`Unable to load texture (${file.name}): ${file.reader.error.message}`, "error");
+						file.reader.abort();
+					};
+
+					file.reader.readAsDataURL(file);
```
Qaomen, line 124: Not ideal, since the data will be saved as base64 string (and increase in size by about 33%), but that's something I'll change myself once I fix `*.glb` support.

src/client/editor/textures.js
Qaomen: There's no check whether the texture is valid on load. I'd use the onLoad/onError callbacks from `TextureLoader.load` before creating the `Texture` object, similar to how it's handled for models.

src/client/editor/materials.js
```diff
@@ -1,39 +1,256 @@
...
+			if (event.target !== uvButton) {
+				return;
+			}
+
+			this.classList.toggle("visible");
+			this.getElementsByClassName("textureUVTransform")[0].classList.toggle("visible");
+		}, false);
+	}
+
+	// Load material-related data if it exists in the project
```
Qaomen: Maybe other loading of project data (like roughness, textures) should be here as well, since it's implied.
It also shouldn't be necessary to check whether properties in `projectData` exist / are valid, since they should be at all times.

src/client/editor/materials.js
```diff
@@ -1,39 +1,256 @@
...
+		this.elements.itemName.value = this.name = this.projectData.name;
+	} else {
+		this.elements.itemName.value = this.name = this.projectData.name = "";
+	}
+
+	// Array of option elements that are part of the model material selection
+	this.optionElements = [];
+
+	// Set a default material
+	this.compiledMaterial = new DefaultMaterial();
```
Qaomen: Already happens on line 11.

src/client/editor/materials.js
```diff
@@ -1,39 +1,256 @@
...
+		selectElement.addEventListener("change", function() {
+			self.projectData[property].textureUuid = this.value;
+		}, false);
+
+		if (typeof this.projectData[property].textureUuid !== "undefined") selectElement.value = this.projectData[property].textureUuid;
+	}
+
+	let setSide = function() { self.projectData.side = this.value; };
+	this.element.getElementsByClassName("side")[0].addEventListener("change", setSide, false);
+	if (typeof this.projectData.side !== "undefined") this.element.getElementsByClassName("side")[0].value = this.projectData.side;
```
Qaomen: As mentioned before, the `undefined` checks can be removed.
The `side` dropdown doesn't have a option selected by default.

src/client/editor/materials.js
```diff
@@ -1,39 +1,256 @@
...
+
+Material.prototype.toggleCollapse = function() {
+	this.element.getElementsByClassName("collapse")[0].children[0].classList.toggle("rotated");
+	this.element.classList.toggle("collapsed");
+};
+
+Material.prototype.createOptionElement = function() {
+	let optionElement = document.createElement("option");
+	optionElement.className = `option-${this.uuid}`;
+	optionElement.innerText = this.name;
```
Qaomen: Should we add the uuid as a prefix here like `[uuid] Name`? We get blank options in the dropdown for empty names otherwise, looks a bit funny.

src/level/level.js
```diff
@@ -20,9 +22,32 @@ function Level() {
 	};
 }
 
+Level.prototype.addTexture = function(uuid, fileContents) {
+	this.textures[uuid] = {
```
Qaomen, line 26: Texture name should be part of the properties here.

src/level/level.js
```diff
@@ -20,9 +22,32 @@ function Level() {
...
+		file: fileContents
+	};
+	return this.textures[uuid];
+};
+
+Level.prototype.addMaterial = function(uuid) {
+	this.materials[uuid] = {
```
Qaomen, line 45: Material name should be part of the properties here. Since the properties of the texture data are known, they could be set to defaults here too.

src/level/level.js
```diff
@@ -20,9 +22,32 @@ function Level() {
...
+		entities: {},
+		"diffuse-a": {},
+		"diffuse-b": {},
+		mask: {},
+		"normal-a": {},
+		"normal-b": {},
+		side: 0, // THREE.FrontSide, but to avoid importing the entire THREE library for just this...
```
Qaomen: (You can't import the THREE library here anyway, this script's used on the server as well!)
There's a type mismatch here. It's stored as an int by default (as it is in THREE), but this is changed to a string when you modify it (`"FrontSide"` `"BackSide"` `"DoubleSide"`).

src/client/editor/materials.js
```diff
@@ -1,39 +1,256 @@
...
+		normalB:  toTextureProperty(this.projectData["normal-b"])
+	};
+
+	try {
+		let customMaterial = new CustomMaterial(properties);
+		this.compiledMaterial.copy(customMaterial.material);
+		this.compiledMaterial.build();
+		this.compiledMaterial.needsUpdate = true;
+	} catch (error) {
+		console.warn(error);
```
Qaomen, line 184: This should also log to `editorLog`.

src/client/render/custom-material.js
```diff
@@ -0,0 +1,94 @@
...
+		);
+	}
+
+	let diffuseNodeA,
+		diffuseNodeB,
+		maskNode,
+		normalNodeA,
+		normalNodeB;
+
+	diffuseNodeA = new THREE.TextureNode(properties.diffuseA.texture);
```
Qaomen: Non-existing properties (like `diffuseA.texture` if nothing is set) aren't handled. This is caught in `Material.parse()` but since roughness/metalness are handled (though as far as I know always have a value), I'd either check all or none of them, or maybe `try/catch` in this function and guarantee a valid returned material.

That said, what should happen if a texture isn't set? Use a generated fallback texture?

src/client/editor/materials.js
```diff
@@ -1,39 +1,256 @@
...
+		diffuseB: toTextureProperty(this.projectData["diffuse-b"]),
+		mask:	  toTextureProperty(this.projectData["mask"]),
+		normalA:  toTextureProperty(this.projectData["normal-a"]),
+		normalB:  toTextureProperty(this.projectData["normal-b"])
+	};
+
+	try {
+		let customMaterial = new CustomMaterial(properties);
+		this.compiledMaterial.copy(customMaterial.material);
+		this.compiledMaterial.build();
```
Qaomen: Removing this line fixes the weird not-getting-updated bug. Just setting `needsUpdate` to true will cause it to rebuild before render. (Apparently it passes the renderer as an argument to this internally or something but either way this seems to do the trick!)

Might be worth re-checking the weird water performance drop after this?

src/client/editor/models.js
```diff
@@ -16,44 +17,101 @@ function Model(name, sceneObject, projectData) {
...
+	this.element.getElementsByClassName("itemName")[0].innerHTML = name;
+	this.element.getElementsByClassName("showItem")[0].addEventListener("mousedown", function() {
 		modelsTab.select(name);
 	}, false);
 
+	// List model children
+	let childMeshes = _getChildMeshes(this.sceneObject);
+	for (let i = 0; i < childMeshes.length; i++) {
+		if (!this.projectData.childMeshes[i]) {
+			this.projectData.childMeshes[i] = {};
```
Qaomen: Project data should have valid properties inside.

src/client/editor/models.js
```diff
@@ -168,6 +226,89 @@ Model.prototype.getConcaveGeometry = function() {
 	return this.concaveGeo;
 };
 
+function ChildMesh(mesh, parent, projectData) {
+	this.mesh = mesh;
+	this.parent = parent;
+	this.projectData = projectData;
+
+	if (!this.projectData) {
```
Qaomen: This should always exist, so I don't think creating a separate untracked object like this will do much good.

src/client/editor/models.js
```diff
@@ -168,6 +226,89 @@ Model.prototype.getConcaveGeometry = function() {
...
+	this.element.appendChild(childElementName);
+	this.element.appendChild(this.selectElement);
+	this.parent.element.getElementsByClassName("itemDetails")[0].appendChild(this.element);
+}
+
+ChildMesh.prototype.setMaterial = function(materialUuid) {
+	// If no material is provided, fall back to the original material.
+	if (!materialUuid) {
+		this.mesh.material = this.originalMaterial;
+		this.projectData.material = null;
```
Qaomen, line 278: Realising that this won't pass validation as it expects a string, and null is considered an object. This is actually an existing bug when setting a model for prefabs too it seems... I'll fix this myself later because this needs some core validation changes maybe. Making a note of it here just in case!

src/client/editor/project.js
```diff
@@ -241,16 +243,33 @@ let projectTab = function() {
...
+			while (Object.keys(texturesTab.textures).length > 0) {
+				texturesTab.removeTexture(Object.keys(texturesTab.textures)[0]);
 			}
 
 			//Initialize loaded project
 			editorLog("Loading project...");
 			this.activeProject = loadedProject;
 
+			for (let key in this.activeProject.textures) {
+				texturesTab.addTexture(key, this.activeProject.textures[key].name, this.activeProject.textures[key].file, this.activeProject.textures[key]);
```
Qaomen: The way it's currently loaded is synchronous and (like mentioned in a comment in `textures.js`) doesn't check for errors. Just a reminder that if that code changes, this would change to something probably similar to how model w/ promises are done.

src/client/render/render-core.js
```diff
@@ -224,7 +225,62 @@ MarbleLevel.prototype.loadLevel = function(data) {
 	this.water.setHeight(data.world.waterLevel);
 	this.sky.recalculate({ inclination: data.world.sunInclination });
 
-	// Load models
+	// Load textures
+	let textures = {};
+	for (let textureUuid in data.textures) {
+		textures[textureUuid] = Object.assign({}, data.textures[textureUuid]);
```
Qaomen: Not sure why this is copied? Could just make `textures[uuid]` a texture directly?

src/client/render/render-core.js
```diff
@@ -224,7 +225,62 @@ MarbleLevel.prototype.loadLevel = function(data) {
...
+			diffuseA:  data.materials[materialUuid]["diffuse-a"],
+			diffuseB:  data.materials[materialUuid]["diffuse-b"],
+			mask:	   data.materials[materialUuid]["mask"],
+			normalA:   data.materials[materialUuid]["normal-a"],
+			normalB:   data.materials[materialUuid]["normal-b"]
+		};
+
+		try {
+			materials[materialUuid] = (new CustomMaterial(properties)).material;
+			materials[materialUuid].build();
```
Qaomen: Like in `Material.parse()`, setting `needsUpdate` to true is enough. (Haven't tested it here specifically but y'know)

src/client/render/render-core.js
```diff
@@ -305,6 +365,24 @@ function Water(sunLight, waterLevel = 0, fog = false) {
 	this.waterObject.rotation.x = -Math.PI / 2;
 	this.waterObject.position.y = waterLevel;
 	this.waterObject.material.uniforms.size.value = 8;
+	let tempFunction = this.waterObject.onBeforeRender;
```
Qaomen: I mean, it's not *that* temporary of a function, right? Maybe use a name that sounds like it's not leftover testing code? :D

src/client/render/render-core.js
```diff
@@ -305,6 +365,24 @@ function Water(sunLight, waterLevel = 0, fog = false) {
...
+	this.waterObject.onBeforeRender = function(renderer, scene, camera) {
+		let tempHide = [];
+		for (let object of parentScene.children) {
+			if (object.userData.reflectInWater !== true && object.visible === true) {
+				tempHide.push(object);
+				object.visible = false;
```
Qaomen, line 402: I'd consider this a hotfix. This code loops over all objects every frame twice, which isn't very performant. It could use a `THREE.Group()` instead. (A different changelist actually added `levelScene` for level objects, which coincidentally could be used here! This branch hasn't been updated to that point yet, though.) This code also still reflects all marbles.

Maybe, if that custom material issue is resolved with the `needsUpdate` thing, this hotfix isn't that necessary.

src/level/level.js
```diff
@@ -10,6 +10,8 @@ function Level() {
 		roundLength: 160
 	};
 
+	this.textures = {};
+	this.materials = {};
```
Qaomen, line 14: Material validation is missing in `Level.validateLevel()`.

src/level/level.js
```diff
@@ -10,6 +10,8 @@ function Level() {
 		roundLength: 160
 	};
 
+	this.textures = {};
```
Qaomen, line 13: Texture validation is missing in `Level.validateLevel()`.

Qaomen (changes requested):
Boy is it a big one indeed! Pretty hype for seeing this pushed live!

There's some changes to be made, some I think got caused by this whole thing being a tough cookie to crack. Don't take the feedback points too harshly; Tough cookies make it hard to code sometimes.
Also, some comments are just comments. Things I noticed I messed up prior or will change later. I decided to add them anyway for sake of clarity. (And to own up to them, hah)

One thing to take note on overall (which is something I should've stated somewhere clearly so this is on me), is that the project data should always be valid and complete at all times. Else, it won't pass validation thus can't be saved. TLDR: Project properties should exist from the beginning. As a result, checking whether they exist isn't necessary.

Also, I personally wouldn't consider #154 to be closed after this. Reasoning is in a comment at the relevant code change.

src/client/editor/textures.js
Zemanzo: but then you'll get `texture.texture` everywhere and that pains me D:

src/client/editor/project.js
Zemanzo: I'm *decently* sure that it will remain synchronous, as the TextureLoader generates a `Texture` and returns it synchronously. It just updates it once it has actually loaded. But thanks for the reminder!

src/client/editor/textures.js
Zemanzo: I thought of this as well. But in the end, you have to juggle 5 selection elements. We could remove and re-add all material options when updating, (which you do with models as well, I think?) but I've never been a fan of the whole remove-and-recreate. The performance impact in this case is minimal and it's easier to maintain so I might end up changing it anyway, but I hope you share my sentiment.

src/client/editor/materials.js
Zemanzo: Does not fix the performance drop with water / custom materials unfortunately...

src/client/render/render-core.js
Zemanzo: The `needsUpdate` does not help with the performance drop. Will leave this hotfix in until a proper fix is implemented in #154

src/client/render/render-core.js
Zemanzo: I'm sure I had a reason for this, but it works fine without so I guess it was a poor reason.

src/client/editor/models.js
Zemanzo: I think in this case it's required to check the `projectData`. When adding a new model, it's impossible to know how many childMeshes there are. Unless you would prefer this section to be in the `modelsTab.loadModel()` function? That would give even more headaches, because `ChildMesh` relies on the parent `Model`...

src/client/editor/textures.js
Zemanzo: Gonna keep it as-is.

src/client/editor/models.js
Qaomen: Sorry for the confusion, I meant how `this.projectData.childMeshes[i]` is set to an empty object in this case. The object would ultimately have a `material` property, but I don't think this is actually set anywhere.

src/client/editor/project.js
Qaomen: I get the feeling the loading of the image is done synchronously as well, since it seems to block the main thread while loading a project (which isn't the case with models I don't think, which also loads textures..?). It's not a huge issue, but we might want to look at it eventually, maybe. Very maybe.

src/client/render/custom-material.js
```diff
@@ -0,0 +1,140 @@
...
+		let normalBlend = new THREE.Math3Node(
+			normalNodeA,
+			normalNodeB,
+			maskAlphaChannel,
+			THREE.Math3Node.MIX
+		);
+
+		this.material.normal = new THREE.NormalMapNode(normalBlend);
+	} else if (properties.normalA) {
+		this.material.normal = new THREE.NormalMapNode(normalNodeA);
```
Qaomen, line 132: If `normalB` isn't set, `normalA` is applied everywhere despite the mask possibly being set. Is it possible to make a blend node with just `normalA`/`normalB` and a `mask`? This could be intentional by the user.

src/client/render/render-core.js
Qaomen: I actually meant assigning `textures[textureUuid]` to the project data in general. Only `textures[textureUuid].texture` is used in the code, so you could do `textures[textureUuid] = new TextureLoader().load(...)` instead and leave the project data out.

src/level/level.js
```diff
@@ -118,6 +130,37 @@ Level.prototype.validateLevel = function() {
 		validateObject(this.world, template.world, true);
 		validateObject(this.gameplay, template.gameplay, true);
 
+		// Validate texture properties
+		if (this.textures) {
```
Qaomen: This property (along with the materials one) should at least exist as an empty object, so this should throw an error if it doesn't.

Zemanzo: Yes, *if* it exists, that is. It's not a vital level property, right?

Zemanzo: So something like `if (this.textures && typeof this.textures === "object") {`

Qaomen (approved):

PR #169, opened 2019-05-24
Fixed losing Discord authorization + some refactoring

### LINKED ISSUE #132: Cancelling Discord authorization is handled incorrectly
**Expected behavior**
When clicking cancel, it should close the pop-up that appeared and possibly warn the user that the action has been cancelled.

**Actual behavior**
The pop-up remains open and navigates to the regular chat window. (Which is unnecessary, since you already have the chat open to get the pop-up in the first place)

**Reproduction steps**
1. Make sure you are currently not logged in
2. Go to `/chat` (or `/client`, which has `/chat` in an iframe)
3. Click the "Log in with Discord" button
3. Cancel the authorization
4. Pop-up remains visible with another copy of the chat

### LINKED ISSUE #144: Opening multiple client instances stops client chat from working
**Expected behavior**
When multiple client instances are opened for the same user, the user should be able to use both instances.

**Actual behavior**
Opening a new instance prevents the user from sending chat messages in older ones.

**Reproduction steps**
1. Open the client in your browser of choice while the server is running.
2. Open another in a new tab or window.
3. Send any chat message in the first client (or press !marble).
4. The message doesn't show up, the server gives a `User ID and access token mismatch!` warning.

**Additional context**
It only has to do with sending chat messages. Receiving them (and everything else) works just fine.

Zemanzo:
Resolves #144 
Resolves #132


src/server/chat/discord-manager.js
```diff
@@ -0,0 +1,178 @@
+const config = require("../config");
+const log = require("../../log");
+const messages = require("./messages");
+const discord = require("discord.js");
+const request = require("request-promise-native");
+
+function DiscordManager() {}
```
Qaomen: Since there will only be one `DiscordManager`, this should be a module.

Qaomen (changes requested):
Working good! Just a minor thing for code consistency.

Qaomen (approved):

PR #183, opened 2019-05-29
Client cleanup and small fixes

### LINKED ISSUE #164: Client clean-up and restructure
**Description**
The client-side scripts have been getting more cluttered over time and it's not always clear what is supposed to go where. Additionally, asynchronous code isn't always handled cleanly, causing race conditions. Some clean-up is in order.

- [x] Asynchronous code should be handled correctly.
- [x] Level managing code should be standalone (currently part of rendering core).
- [x] Script inter-dependencies should be cleared up and/or resolved.
- [ ] An overview on the architecture and loading dependencies should be available in the TDD.

Qaomen:
**Overview**
- Level Management is split from `renderCore`, though it's just a point of access for now.
- Level data helper function module thingy is now called `levelIO`.
- `client.js` handles module initialization so modules actually exist before they're accessed by each other. (Resolves the inter-dependency issue)
  - `networking` is initialized synchronously, and waits on `game` for message processing.
  - `renderCore` is initialized synchronously: Fallback model is loaded async (replaces a sync-loaded fallback) and viewport added to DOM when ready. It's not necessary to wait on these explicitly.
  - `game` is initialized when DOM is ready, and returns a Promise that resolves once initialization is complete.
- Cleaned up `networking.js`.
- Handling of initial game data is all done in `game.initializeGameState()`.

**Smaller stuff**
- Fixed an error when a HUD notification is removed.
- Temporarily removed `*.glb` support.
- Water's `onBeforeRender` now disables the `levelObjects` scene instead of looping through all current objects.
- Added a default case for network message types.

**Referencing issues**
Closes #164


src/client/client/game.js
```diff
@@ -280,17 +290,35 @@ let game = (function() {
...
+			// Spawn marbles
+			for (let i = 0; i < gameState.initialMarbleData.length; i++) {
+				this.spawnMarble(gameState.initialMarbleData[i]);
+			}
 
-				this.setCurrentGameState({state: _serverData.currentGameState}, true);
-			});
+			// Start loading the level asynchronously
+			let levelName = gameState.levelId;
+			fetch(`/resources/maps/${levelName}.mmc`)
```
Zemanzo, line 306: ~~It might be a good idea to already do this request in the `initialize` function, store it in a variable and use that here, so it starts downloading right away, but does not start processing it yet.~~

src/client/editor/editor.js
```diff
@@ -226,9 +229,20 @@ let editor = function() {
...
+let gridHelper = new THREE.GridHelper(20, 20);
+levelManager.activeLevel.scene.add(gridHelper);
+gridHelper.position.y = -.01;
+let axesHelper = new THREE.AxesHelper(3);
+levelManager.activeLevel.scene.add(axesHelper);
+
+// Initialize DOM elements and tabs
+domReady.then( () => {editor.initialize();} );
 
 window.onbeforeunload = function(e) {
```
Zemanzo: Kind of unrelated to this PR but I suddenly noticed this. Should we make a rule about these kind of event listeners? I prefer using `.addEventListener()` over `.onEvent = function`, because then you're certain you're not overwriting any formerly set events.

src/client/level-manager.js
```diff
@@ -0,0 +1,317 @@
...
+			if (this.activeLevel) {
+				renderCore.getMainScene().remove(this.activeLevel.scene);
+			}
+			this.activeLevel = marbleLevel;
+			renderCore.getMainScene().add(this.activeLevel.scene);
+		},
+
+		initialize: function() {
+			let level = new MarbleLevel();
+			level.loadLevel(new Level()); // Load default level properties
```
Zemanzo: This `MarbleLevel` and `Level` thing is a little confusing. I mean, quickly enough I got what was done here, but I figured I would bring it up regardless. Might be worth a small discussion for renaming?

Zemanzo (changes requested):

src/client/editor/editor.js
Qaomen: I agree, we should make a rule of that.

src/client/client/game.js
Zemanzo: Yeah that was a brainderp and a half.

Zemanzo (approved):

Issue #185, opened 2019-06-03
Free camera helper UI


Zemanzo:
**Description**
The free camera has quite some features, but how to use it is not well explained to the user.

**Minimum Acceptance Criteria**
- [ ] When the free camera is active for the first time, show a graphic that explains how to control the camera. (Click & hold; WASD + QE)
- [ ] Once they have successfully moved the camera, remove the graphic.
- [ ] Display speed changes on screen

**Is blocked by the following issues**
#31 
Probably best if #81 is also done before this one

PR #186, opened 2019-06-03
Free camera improvements

### LINKED ISSUE #31: Free camera
- [x] Camera can be freely controlled using the arrow keys / WASD
- [x] Camera can be moved up/down (relative) using the Q and E keys.
- [x] Camera can be sped up & slowed down with scrolling.     

**Related issues**
Switching cameras is described in #81

Zemanzo:
Some code cleanup done.
Changed default position back to view the current map starting area.
Users can now scroll to speed up or slow down the camera movement speed.
Users can use Q and E to move down and up (relative to the camera angle).

Closes #31 

This does **NOT** yet add any helper UI, it will be done in a separate issue ( #185 )

### DIFF
--- a/src/client/render/cameras.js
+++ b/src/client/render/cameras.js
@@ -196,7 +196,6 @@ function FreeCamera(
 
 	if (options.disableOnBlur === true) {
 		document.addEventListener("visibilitychange", (function() {
-			console.log(document.hidden);
 			if (document.hidden) {
 				this.disable();
 			} else {
--- a/src/client/render/cameras.js
+++ b/src/client/render/cameras.js
@@ -27,7 +27,7 @@ let addRegisteredEventListener = function(scope, event, func, capture) {
 	@method disable Disables the controls.
 	@method update Updates the controls. Should be called in some update loop.
 */
-export function CameraFlyControls(
+function FreeCamera(
 	scene,
 	renderer,
 	options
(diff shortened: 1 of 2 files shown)


src/client/render/cameras.js
```diff
@@ -156,70 +196,68 @@ export function CameraFlyControls(
 
 	if (options.disableOnBlur === true) {
 		document.addEventListener("visibilitychange", (function() {
-			if (document.hidden)
+			console.log(document.hidden);
```
Zemanzo: Oops I will remove this

src/client/render/cameras.js
```diff
@@ -80,21 +101,32 @@ export function CameraFlyControls(
...
-		listeners.push( addRegisteredEventListener(options.pointerLockElement, "mousedown", func, false) );
+		_listeners.push( addRegisteredEventListener(options.pointerLockElement, "mousedown", func, false) );
 
 		// Release pointerlock
 		func = function() {
-			controls.unlock();
+			self.controls.unlock();
 		};
-		listeners.push( addRegisteredEventListener(document, "mouseup", func, false) );
+		_listeners.push( addRegisteredEventListener(document, "mouseup", func, false) );
```
Qaomen, line 118: Some of these event listeners could be a bit cleaner. This also works, without needing to set up `func` or using `self`:
`_listeners.push( addRegisteredEventListener(document, "mouseup", () => {this.controls.unlock();}, false) );`
I guess it's personal preference in the end though.

Qaomen (changes requested):
Good stuff! Just that logging that needs removing and a style thing that I'll leave up to you whether you want to change it (and to what extent).

src/client/render/cameras.js
Zemanzo: Not going to bother. I kinda like `func` here because it allows the `push` to `_listeners` to be on a single line. Good suggestion though (:

Qaomen (approved):

PR #189, opened 2019-06-03
Marble manager, deltatime fixes, configurable reflections

### LINKED ISSUE #154: Improve water reflection

### LINKED ISSUE #161: Marble Manager
**Description**
Since marbles will eventually be used by both the client and editor, and needs to be updated before rendering, we could do with a manager that keeps track of all marbles in the scene that updates/interpolates them based on input by the client (networking) or editor (physics). Lots of the final design is still subject to change.
This will also remove the marble update code from the renderer.

- [x] A `MarbleManager` module that should exist in `src/client/`
- [x] Should have functions to add/remove a single marble, remove all marbles.
- [ ] Should have an interpolate marbles function for the renderer to use.
- [x] Should have an entry point to update marble transform data for the networking/physics to use.

Blocked by #69

Qaomen:
I accidentally added the water reflection changes to this branch, so that's what this PR is about now, too!

**PR part the first:**
Adds the marble manager module, which manages any marbles that are active in the scene. Also changed the update process so only one delta time is calculated and passed to the functions that need them. Fixed a camera bug that occurs when the delta time is too high.
(I left the network's marble interpolation as is, since that will be revamped soon enough anyway.)

**PR part the second:**
Removed marble name reflections and added config options for level/marble reflections. I wrapped these in `renderCore` functions since I'd imagine these can be changed later as graphics settings.

**Referencing issues**
Closes #161
Closes #154


src/client/marble-manager.js
```diff
@@ -0,0 +1,135 @@
...
+import * as Cookies from "js-cookie";
+
+let _userData = Cookies.getJSON("user_data");
+
+// This module manages all the marbles that physically exist in the scene.
+let marbleManager = function() {
+	let _marbles = []; // Array of marbles that currently exist in the scene
+	return {
+		marbleGroup: null, // Group containing marble instances
+		marbleNames: null, // Group containing name sprites
```
Zemanzo: Maybe call this `marbleNamesGroup` for consistency? Up to you.

src/client/render/render-core.js
```diff
@@ -138,114 +139,28 @@ let renderCore = function() {
...
-				);
-
-				// Also update the nameSprite position
-				if (_marbleMeshes[i].nameSprite) {
-					_marbleMeshes[i].nameSprite.position.x = (_marbleMeshes[i].mesh.position.x || 0);
-					_marbleMeshes[i].nameSprite.position.y = (_marbleMeshes[i].mesh.position.y || 0) + _marbleMeshes[i].size - .1;
-					_marbleMeshes[i].nameSprite.position.z = (_marbleMeshes[i].mesh.position.z || 0);
-				}
-			}
+		getMainScene: function() {
```
Zemanzo: Not entirely clear on why this is a function? I understand that for some properties it would be nice to have them as a getter function (due to changes that will be made later), but for the main scene I don't think that is necessary? Nor is it for the default model probably. Why not just make these public variables?

src/client/editor/editor.js
```diff
@@ -126,6 +127,8 @@ let editor = function() {
 			prefabsTab.initialize();
 			worldTab.initialize();
 
+			renderCore.clientUpdateCallback = this.update;
```
Zemanzo: Setting the **client**UpdateCallback in the **editor** is a bit strange, isn't it?

src/client/level-manager.js
```diff
@@ -212,18 +213,24 @@ function Water(parent, sunLight, waterLevel = 0, fog = false) {
 	this.waterObject.material.uniforms.size.value = 8;
 	let originalOnBeforeRender = this.waterObject.onBeforeRender;
 	this.waterObject.onBeforeRender = function(renderer, scene, camera) {
-		parent.levelObjects.visible = false;
+		if(!renderCore.waterReflectsLevel()) parent.levelObjects.visible = false;
+		if(!renderCore.waterReflectsMarbles()) marbleManager.marbleGroup.visible = false;
```
Zemanzo, line 217: I think it might be slightly nicer to have a `=== false` (or `!== true`, whatever floats your boat) check here, and it's faster because it doesn't have to do any type conversions!

Zemanzo (commented):
Couple of things that aren't entirely clear to me, but that might be just because I'm a dum. I skipped over the network stuff because it will be changed later anyway.

src/client/editor/editor.js
Qaomen: Totally is, renamed the callback to `updateCallback`.

src/client/render/render-core.js
Qaomen: I agree there's no need for it for `getMainScene`, so I'll change that. The default model had a getter because at one point it might've involved a Promise due to it loading dynamically, but that's been solved a different way for now.

Zemanzo (approved):

PR #192, opened 2019-06-05
Tracking camera

### LINKED ISSUE #32: Tracking camera
**Description**
A camera that automatically follows marbles to avoid the need to manually control the camera.

- [x] Used by default, for own marble.
- [x] Players can select marble that the camera should track by clicking the corresponding user in the list on the right.
- [x] Camera automatically tracks the selected marble.
- [x] Camera follows it whilst it rolls down the track.
- [x] When the selected player does not have a marble to track, it should move back to the default camera position (usually the start of a level).

**Related issues**
Switching cameras is described in #81

Zemanzo:
The new default camera for the client. It tracks your own marble by default, but you can click other marbles in the list to have the camera track those instead. When there is nothing to track, it will move back to its default position.

I also tweaked the marble entries list a bit, so it looks nicer and to allow for the camera tracking button.

**Referencing issues**
Closes #32

### DIFF
--- a/src/client/client/game.js
+++ b/src/client/client/game.js
@@ -290,7 +290,7 @@ let game = function() {
 			_enteredMarbleList[marble.entryId] = marble;
 
 			// Add mesh
-			marble = marbleManager.spawnMarble(marble);
+			marbleManager.spawnMarble(marble);
 
 			// Add UI stuff
 			let listEntry = _DOMElements.marbleListTemplate.cloneNode(true);
--- a/src/client/client/game.js
+++ b/src/client/client/game.js
@@ -39,17 +39,18 @@ let game = function() {
 		_requestComplete,
 		_requestStart = Date.now();
 
-	let _toggleMarbleTracking = function(marble, mesh) {
+	let _toggleMarbleTracking = function(marble) {
 		if (renderCore.controls.type === "TrackingCamera") {
+			let mesh = null;
 			if (_marbleBeingTracked === marble) {
 				_marbleBeingTracked = null;
-				mesh = null;
 				marble.listEntryElement.getElementsByClassName("camera")[0].classList.remove("selected");
 			} else {
 				if (_marbleBeingTracked) {
 					_marbleBeingTracked.listEntryElement.getElementsByClassName("camera")[0].classList.remove("selected");
 				}
 				_marbleBeingTracked = marble;
+				mesh = marble.mesh;
 				marble.listEntryElement.getElementsByClassName("camera")[0].classList.add("selected");
 			}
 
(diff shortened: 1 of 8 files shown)


src/client/client/game.js
```diff
@@ -263,18 +289,30 @@ let game = function() {
...
+			listEntry.getElementsByClassName("name")[0].innerText = marble.name;
 			listEntry.getElementsByClassName("name")[0].innerText = marble.name;
 			listEntry.getElementsByClassName("color")[0].style.background = marble.color;
-			listEntry.getElementsByClassName("time")[0].innerText = marble.time ? `🏁 ${(marble.time * .001).toFixed(2)}s` : "";
+			listEntry.getElementsByClassName("time")[0].innerText = marble.time ? `${(marble.time * .001).toFixed(2)}s` : "";
 			listEntry.getElementsByClassName("rank")[0].innerText = !isNaN(marble.rank) && marble.rank !== null ? `#${marble.rank + 1}` : "";
 			listEntry.style.order = marble.rank;
 			_enteredMarbleList[marble.entryId].listEntryElement = listEntry;
 
+			if (_userData && _userData.username === marble.name) {
```
Qaomen: Checking for `_userData.id === marble.userId` is probably the safer option here.

src/client/client/game.js
```diff
@@ -263,18 +289,30 @@ let game = function() {
 			_enteredMarbleList[marble.entryId] = marble;
 
 			// Add mesh
-			marbleManager.spawnMarble(marble);
+			let mesh = marbleManager.spawnMarble(marble);
```
Qaomen: Maybe we should make the mesh a property of the marble object that is stored in `_enteredMarbleList`. It could be useful in the future (like accessing oob marbles directly?), and `_toggleMarbleTracking` will only need one argument which is a little cleaner.

src/client/render/render-core.js
```diff
@@ -23,8 +22,8 @@ let renderCore = function() {
 		_previousTime = now;
 
 		// Update active controls, needs to be buttery smooth, thus is called before requesting the next frame
-		if (_controls.enabled === true) {
-			_controls.update(deltaTime);
+		if (renderCore.controls.enabled === true) {
+			renderCore.controls.update(deltaTime);
```
Qaomen: If we move this after the `updateCallback`, the tracking camera will use the updated marble positions.

Qaomen (changes requested):
Liking the entries list tweaks, loving the tracking camera! There's a small request regarding the marble+mesh and a reminder on the update order, but other than that all's fine.

The deltaTime in the tracking camera is something we might address later.

Qaomen (approved):

PR #196, opened 2019-06-10
Marble name improvements

### LINKED ISSUE #165: Marble name improvements
**Description**
In-game, marbles have name sprites above them so you can easily identify whose marble it is, but the names can disappear behind geometry or be unreadable from a distance. With some changes we can improve the experience.

- [ ] Name sprites shouldn't use depth tests or write to the depth buffer (aka should be visible through level geometry).
- [ ] Names should always be the same size on-screen. If this doesn't turn out well in practice, have depth influence the size but with a minimum size limit.
- [ ] The client's marble (the one currently displayed in a different colour) should always be on top.

Sanquinary:
**Title**
Sprite improvements for marbles

**Referencing issues**
Closes #165

### DIFF
--- a/src/client/marble-manager.js
+++ b/src/client/marble-manager.js
@@ -126,11 +126,11 @@ const makeTextSprite = function(message, options = {}) {
 	let texture = new THREE.Texture(canvas);
 	texture.needsUpdate = true;
 
-	let spriteMaterial = new THREE.SpriteMaterial({ 
+	let spriteMaterial = new THREE.SpriteMaterial({
 		map: texture,
 		sizeAttenuation: false,
 		depthWrite: false,
-		depthTest: false,
+		depthTest: false
 	});
 
 	let sprite = new THREE.Sprite(spriteMaterial);
--- a/src/client/marble-manager.js
+++ b/src/client/marble-manager.js
@@ -101,6 +101,7 @@ const MarbleMesh = function(marbleData) {
 	let nameSpriteOptions = {};
 	if (_userData && _userData.username === this.name) {
 		nameSpriteOptions.color = "#BA0069";
+		nameSpriteOptions.drawOnTop = 9e9;
 	}
 
 	// Add name sprite (we avoid parenting, because this will also cause it to inherit the rotation which we do not want)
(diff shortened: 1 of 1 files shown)


src/client/marble-manager.js
```diff
@@ -101,6 +101,7 @@ const MarbleMesh = function(marbleData) {
 	let nameSpriteOptions = {};
 	if (_userData && _userData.username === this.name) {
 		nameSpriteOptions.color = "#BA0069";
+		nameSpriteOptions.drawOnTop = 9e9;
```
Zemanzo: I like that you made this a variable in the `nameSpriteOptions`, but the variable name does not really represent what it actually does. I would suggest renaming this to `nameSpriteOptions.renderOrder`

src/client/marble-manager.js
```diff
@@ -125,9 +126,16 @@ const makeTextSprite = function(message, options = {}) {
...
+	let spriteMaterial = new THREE.SpriteMaterial({ 
+		map: texture,
+		sizeAttenuation: false,
+		depthWrite: false,
+		depthTest: false,
+	});
+
 	let sprite = new THREE.Sprite(spriteMaterial);
-	sprite.scale.set(4, 1, 1.0);
+	sprite.scale.set(0.65, 0.25, 1.0);
```
Zemanzo: This results in a stretched looking name. It's also a tad big as it obfuscates a lot of the marbles and scenery. It should be large enough to be legible, but not much bigger than that.

It would be nice to have the size configurable, but that is out of the scope of this PR / issue.

src/client/marble-manager.js
```diff
@@ -125,9 +126,16 @@ const makeTextSprite = function(message, options = {}) {
...
+	sprite.renderOrder = options.drawOnTop || 1;
```
Zemanzo: Why bother setting the renderOrder if no custom one is defined? It's better to leave it at the default. I would suggest something more like `if (options.drawOnTop) sprite.renderOrder = options.drawOnTop;` (but with a better name, as mentioned in an earlier comment)

Zemanzo (changes requested):
Functionally, it works great! However, there are some minor improvements to be made. Awesome to have you contribute though, thank you!

Zemanzo (commented):
Changes look good! I will hold off of approving this until @Qaomen has taken a look at it though.

src/client/marble-manager.js
```diff
@@ -125,9 +126,16 @@ const makeTextSprite = function(message, options = {}) {
 	let texture = new THREE.Texture(canvas);
 	texture.needsUpdate = true;
 
-	let spriteMaterial = new THREE.SpriteMaterial({ map: texture });
+	let spriteMaterial = new THREE.SpriteMaterial({ 
+		map: texture,
+		sizeAttenuation: false,
+		depthWrite: false,
+		depthTest: false,
```
Qaomen: This line has a trailing comma which can be removed.

src/client/marble-manager.js
```diff
@@ -125,9 +126,16 @@ const makeTextSprite = function(message, options = {}) {
 	let texture = new THREE.Texture(canvas);
 	texture.needsUpdate = true;
 
-	let spriteMaterial = new THREE.SpriteMaterial({ map: texture });
+	let spriteMaterial = new THREE.SpriteMaterial({
```
Qaomen: This line has a trailing space that can be removed.

Qaomen (changes requested):
Overall, good changes! There are two small warnings that showed up for me with ESLint though, so once these are resolved it's good to go!

Zemanzo (approved):

Qaomen (approved):

PR #205, opened 2019-06-12
Async level loading + few tweaks

### LINKED ISSUE #191: Level loading in worker
**Description**
Loading the level has a significant impact on performance, and halts the main thread. To improve this, it should be handled in a worker process.
Profiling result of 3.77 seconds to decode the MessagePack data alone

- [x] Loading the level is done in a worker as much as possible.
- [x] It should impact the main thread minimally.

**Related issues**
It might be nice to take #173 into account when designing this.

Qaomen:
**Changes**
- `MarbleLevel` now loads default level data in the constructor (instead of being manually set before).
- Added a function that uses a worker thread to fetch/parse level data, and loads the level on the main thread afterwards.
- Level data loading has a bit more error checking now.
- `game.js` will keep track of what level is currently loaded, so it won't unnecessarily reload on reconnect.

On my machine, the async loading part takes about 1600ms which is now off the main thread. Pretty neat!

**Referencing issues**
Closes #191

### DIFF
--- a/src/client/client/game.js
+++ b/src/client/client/game.js
@@ -3,7 +3,6 @@ import * as Cookies from "js-cookie";
 import { levelManager } from "../level-manager";
 import { marbleManager } from "../marble-manager";
 import { renderCore } from "../render/render-core";
-import * as levelIO from "../../level/level-io";
 
 let _userData = Cookies.getJSON("user_data");
 
@@ -14,6 +13,8 @@ let game = function() {
 		},
 
 		_enteredMarbleList = [],
+
+		_currentLevelId = null,
 
 		_serverData = {
 			currentGameState: null,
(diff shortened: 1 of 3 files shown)


src/client/client/game.js
```diff
@@ -343,20 +344,14 @@ let game = function() {
 
 			// Start loading the level asynchronously
 			let levelName = gameState.levelId;
```
Zemanzo: Why do we have this extra variable? Seems a bit unnecessary. I mean, it was probably me who wrote it this way, but that's silly.

src/client/level-loader.worker.js
```diff
@@ -0,0 +1,37 @@
...
+		.then((buffer) => {
+			if(buffer) {
+				let levelData = levelIO.load(buffer);
+				if(levelData === null) {
+					loadingFinished(false, "Failed to load level data.");
+				} else {
+					loadingFinished(true, levelData);
+				}
+			}
+		}).catch( (error) => {
```
Zemanzo: Slightly inconsistent with newlines / no newlines here.

Zemanzo (changes requested):
Just some minor nagging. It works nicely! Shame msgpack is such a slowpoke with the stringy stuff though.

Zemanzo (approved):

PR #208, opened 2019-06-12
v1.1.1

Zemanzo:
Changes:
- Fix crash due to Discord webhook being rate-limited.
- Fix name highlighting when the user has a nickname on the Discord server.


Qaomen (approved):
Almost forgot the highlight fix, good thing you checked!

PR #209, opened 2019-06-12
Toggle camera style

### LINKED ISSUE #81: Ability to toggle between cameras

Zemanzo:
Adds buttons in the top left of the viewport, so players can switch to their preferred camera style.

DOM remains as a partial so it can be easily added to the editor DOM later.

Closes #81


src/client/render/render-core.js
```diff
@@ -148,6 +148,54 @@ let renderCore = function() {
 			}
 		},
 
+		setCameraStyle: function(type) {
+			// If this function is called from an event, use the event origin node to determine what type to become
+			if (this && typeof type !== "string") type = this.dataset.type;
```
Qaomen: I think it's best to prevent the use of `data-*`. They're easy to forget when code and/or parameters change, and it requires additional code like this here.
It would be cleaner to call `this.setCameraStyle("freeCamera")` directly in the event listeners above.

src/client/render/render-core.js
```diff
@@ -148,6 +148,54 @@ let renderCore = function() {
...
+
+			// Helper function that copies position and rotation from previously used camera
+			function copyPositionAndRotation(target, source) {
+				target.camera.position.copy(source.camera.position);
+				target.camera.rotation.copy(source.camera.rotation);
+			}
+
+			// Copy over transform data, disable previously used camera / controls, enable new camera / controls
+			switch (type) {
+			case "TrackingCamera":
```
Qaomen: I realise this was part of the code before, but shouldn't these parameters be without the first capital letter? E.g. `trackingCamera`.

templates/partials/cameraControls.mustache
```diff
@@ -0,0 +1,15 @@
+<div id="cameraControls">
+	<div class="header">
+		<i class="icon-videocam"></i>Camera style:
+	</div>
+	<div class="options">
+		<div id="cameraFree" data-type="FreeCamera">
```
Qaomen: I don't know if it's *that* worth changing, but maybe we shouldn't use `cameraFree` and `freeCamera` interchangeably. It's a bit confusing.

src/client/render/cameras.js
```diff
@@ -288,6 +305,9 @@ function TrackingCamera(
 			75, renderer.domElement.clientWidth / renderer.domElement.clientHeight, 0.1, 5000
 		);
 
+	if (!options.enabledByDefault)
```
Qaomen: This check would set `enabledByDefault` to true if the option is explicitly set to false.

src/client/render/render-core.js
```diff
@@ -59,6 +59,8 @@ let renderCore = function() {
 	return {
 		mainScene: null,
 		controls: null,
```
Qaomen: At this point, it makes more sense to rename this to `activeCamera`.

src/client/render/render-core.js
```diff
@@ -148,6 +148,54 @@ let renderCore = function() {
...
+				break;
+			}
+
+			renderCore.controls.enable();
+
+			if (_viewport) _onCanvasResize();
+
+			// Set selected camera style in DOM
+			domReady.then(() => {
+				let node = document.querySelector(`[data-type=${type}]`);
```
Qaomen: Maybe a redundant comment, but if the element's `id` and camera's argument name match (e.g. `freeCamera` for both), then `getElementById` should work here.

src/client/client/game.js
```diff
@@ -39,7 +39,10 @@ let game = function() {
 		_requestComplete,
 		_requestStart = Date.now();
 
-	let _toggleMarbleTracking = function(marble) {
+	let _toggleMarbleTracking = function(marble, forceTracking) {
```
Qaomen: It's a hindsight, but since this is not a toggle function (like an on/off switch), I think `_trackMarble` makes more sense for the name.

Qaomen (changes requested):
Good feature, but some things that could do with changes, some of them probably should've been caught/discussed sooner.

Another hindsight, it seems to me that the camera's should be using inheritance. They share a good amount of functionality/options, after all.

templates/partials/cameraControls.mustache
Zemanzo: I see your point, but in this case I wanted to stay consistent with the rest of the DOM. This section is all about cameras, so I figured I would prefix all IDs with `camera`
Maybe `cameraOptionFree` and `cameraOptionTracking` as a less confusing alternative?

src/client/render/cameras.js
Zemanzo: Heck, how did I manage to do this for one of the cameras but not the other

src/client/render/render-core.js
Zemanzo: I think we could do even better, and approach it the same way as other libraries using constants. How about define the constants in `cameras.js` such as:
```js
cameras.FREE_CAMERA = 1;
cameras.TRACKING_CAMERA = 2;
```
which can then be checked (say, in `render-core.js`) as
```js
import { cameras, FreeCamera, TrackingCamera } from "./cameras";

...

switch (renderCore.activeCamera.type) {
case cameras.TRACKING_CAMERA:
    ...
    break;
}
```
?

src/client/render/render-core.js
Zemanzo: Agreed, but I'm not sure how to get the DOM node otherwise if I don't change the node ID to the exact type.

Zemanzo:
@Qaomen Whilst we're discussing this anyway, should we move stuff like `setCameraStyle` over to `cameras.js` as well? Seems like it would fit better there, but I don't know if that will make things more complex.

src/client/render/render-core.js
Qaomen: There's a switch case that could be used to set a local string variable, that it'll use to update the correct node for afterwards.

templates/partials/cameraControls.mustache
Qaomen: I see your point. Since we'll switch to proper constants, the confusion won't be there as much. Maybe just leave it as is in this case?

Qaomen:
> 
> 
> @Qaomen Whilst we're discussing this anyway, should we move stuff like `setCameraStyle` over to `cameras.js` as well? Seems like it would fit better there, but I don't know if that will make things more complex.

I think it's best to leave the function in `renderCore`, since it's the renderer that owns/uses the camera.

Qaomen (approved):

PR #210, opened 2019-06-13
Added default marble with proper topology and UVs

### LINKED ISSUE #170: Marble topology / UVs
**Description**
We should have a marble of an adequate poly count that should support textures without stretching them.

* [x] Carefully considered polycount vs sphericalness
* [x] A defined marble topology (Icosphere?)
* [x] Support for textures on marbles, wrapping nicely and seamless.
* [x] A default texture that will be applied to marbles.

Zemanzo:
Closes #170


Qaomen (changes requested):
Good to see that it works, and liking the new material too! Two things though:

Do we really need `marble-default.gltf`? Since we keep the topology simple, there doesn't seem to be any difference from the generated one at the moment.
Things might change in the upcoming marble skins feature, but maybe we should have a default texture simply named `default.png` for now.

Zemanzo:
> Do we really need `marble-default.gltf`? Since we keep the topology simple, there doesn't seem to be any difference from the generated one at the moment.

The main reason I opted for an external model is so we could control the UVs easier. However, when testing if you actually properly implemented rotations in your PR, I noticed the UVs are pretty much identical with the generated one. I will take a look and see if it's indeed good enough.

> Things might change in the upcoming marble skins feature, but maybe we should have a default texture simply named `default.png` for now.

Yeah, it will be changed (already is in that branch) in the next PR that I will make.

Zemanzo:
Closing this since changes are also present in #216

PR #211, opened 2019-06-14
Gameplay network refactor/buffering

### LINKED ISSUE #11: Sync & lerp marble rotations
**Description**
Rotations of marbles rolling down are visualized and interpolated between ticks.

Marbles visually rotate as expected (naturally, without stutters)

**Additional notes**
Consider converting quaternions to long/lat values for reduced data transfer per tick per marble.

### LINKED ISSUE #12: Buffer networking (server & client)
**Description**
A gameplay network buffer with timestamps.

Negate some network problems for the client, such as lag, throttling, dropped packets etc.
Should delay realtime by no more than 2 seconds.

Qaomen:
Oof.

**Changes**
- Networking now uses MessagePack.
- Server sends game updates at a constant rate.
  - Update data has been changed/reduced.
  - HUD notifications are sent immediately. (This is part hotfix(?), part convenience)
- Client maintains an update buffer that it progresses/interpolates through.
  - If the client's buffer runs out (e.g. due to connection issues), the minimum size increases temporarily.
  - Interpolation has been improved.
- Physics no longer simulates if it doesn't need to.
- Minor changes/fixes/renamings in `client/game.js`.
  - Both the server and client `game.js` now use enums instead of strings.
- Added level author and name to `MarbleLevel` properties on load.
- Fixed sockets not setting the `isBinary` flag.

Currently the marble enter countdown doesn't take ping+buffer length into account, meaning a client could try to enter too late because of this delay. But since it's not part of the relevant issues (and already was a minor issue before), this hasn't been resolved yet.

Sometimes, the marble interpolation doesn't appear fluent while the tracking camera is active if the server's tick rate is high. Since it's not apparent with the free camera, it might be the tracking camera. But that's an issue for another time.

**Referencing issues**
Closes #11 
Closes #12


src/client/client/game.js
```diff
@@ -126,18 +117,53 @@ let game = function() {
...
+			case gameConstants.STATE_STARTING:
+				gameInfoClass = "starting";
+				break;
+			case gameConstants.STATE_STARTED:
+				gameInfoClass = "started";
+				break;
+			case gameConstants.STATE_FINISHED:
+				gameInfoClass = "finished";
+				break;
+			}
```
Zemanzo: This switch case does not have a `default` case. It might be nice to add a warning for unknown states, but I'll leave that up to you.

src/client/client/networking.js
```diff
@@ -1,67 +1,94 @@
...
-				break;
-			case "notification":
-				new HUDNotification(message.content, message.duration, message.style);
-				break;
-			default:
-				console.warn(`Received unknown network message of type "${type}".`, message);
-				break;
+		if(typeof event.data === "string") {
+			// This should only be a HUD Notification
+			console.log(event.data);
```
Zemanzo: Left over debug log

src/server/game.js
```diff
@@ -48,10 +49,25 @@ let game = function() {
 
 		_round = null;
 
+	let _netUpdateHandle = null; // Handle to setTimeout
+	let _netInterval = isNaN(config.network.tickRate) ? 50 : 1000 / config.network.tickRate;
```
Zemanzo: Have this default to 100ms instead of 50. Might also be a bit more clear to write it as 1000 / 10? Slightly confused me at first, but that might just be me.

src/server/game.js
```diff
@@ -82,24 +98,167 @@ let game = function() {
...
+
+		// Clear netGameUpdate
+		_netGameUpdate = {};
+
+		if(_marbles.length > 0) {
+			// Set timer for next network update
+			// If we're behind, shorten the interval time
+			let currentDelay = 0;
+			if(combinedDelay === undefined) {
+				// Start of update loop, start tracking accumalted delay
```
Zemanzo: Small typo, it's `accumulated`, you did type it correctly a few lines down :D

src/server/game.js
```diff
@@ -82,24 +98,167 @@ let game = function() {
 		) + config.marbles.scoring.pointsAwardedForFinishing; // Plus potential bonus for finishing
 	};
 
+	// Triggers a network update loop if it hasn't started yet
+	let _triggerNetworkUpdate = function() {
+		if(!_netUpdateHandle) {
+			console.log("Starting netUpdate loop.");
```
Zemanzo: No need to log this, I think?

src/server/game.js
```diff
@@ -82,24 +98,167 @@ let game = function() {
...
+				currentDelay = delta - _netInterval;
+				combinedDelay += currentDelay;
+				if(combinedDelay > _netInterval * 10) {
+					log.warn("Game networking can't keep up! Skipping 10 network ticks...");
+					combinedDelay -= _netInterval * 10;
+				}
+			}
+			_netUpdateHandle = setTimeout( _netUpdate, Math.min(_netInterval, _netInterval - combinedDelay), now, combinedDelay);
+		} else {
+			console.log("No marbles present, stopping updates.");
```
Zemanzo: No need to log this either?

Zemanzo (changes requested):
I wish I could code like this. Seriously, this is amazing. It works very well, the code is very readable and well structured, and you busted this thing out in only a few days.

Just a couple of very tiny things, a typo and some debug logging. Hats off to you sir, I cannot wait to merge this.

src/client/client/game.js
Qaomen: I just noticed that it's a switch case right above another switch case that handles the same parameter, so I'll set that element's class in there and add a default for the main switch case instead.

Zemanzo (approved):

PR #223, opened 2019-06-18
Skins preview & minor improvements

Zemanzo:
Adds a page where you can preview skins. It also adds a couple of new skins, and improves the 2D representation in front of player's names.

Most of this code will be altered soon after this PR, since I would like to redo the skins system. I will document the new design in the TDD.


src/client/skins.js
```diff
@@ -0,0 +1,76 @@
...
+levelManager.initialize();
+marbleManager.initialize();
+
+// Add marbles
+let _rotateMarbles = false;
+let _marble = {
+	skinId: "default",
+	color: "#ffffff"
+};
+_marble.geometry = new THREE.SphereBufferGeometry(1, 32, 32);
```
Qaomen, line 32: At a glace, I think it would be better to use `marbleManager.spawnMarble()` here and let that return the mesh. If we manage the preview marble through the same methods as the client, we'll know for sure they show up the same way. If anything changes in there (e.g. materials or maybe even marble topology), we won't have to edit it here as well.

Qaomen (changes requested):
This is actually a pretty nice unplanned feature!
Since most of this code is subject to change, it might not be worth fixing the thing I mentioned in the comments. Let me know what you think.

src/client/skins.js
Zemanzo: This is a good idea, I'll pour it into an issue and fix it once I get to redoing the skins! I don't think it's worth changing right now though.

Qaomen (approved):

PR #235, opened 2019-06-20
Updated dependencies

Zemanzo:
- Removed dependencies for tests (for now)
- Minor migration change for three.js (QuickHull -> ConvexHull)

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
   ],
   "license": "GPL-3.0-only",
   "dependencies": {
+    "@discordjs/uws": "10.149.0",
     "ammo.js": "kripken/ammo.js#ac08f42d3995cfa2b719aa8a9bb5b766c7f63404",
     "better-sqlite3": "^5.0.1",
     "body-parser": "^1.18.2",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -48,9 +49,8 @@
     "request-promise-native": "^1.0.7",
     "semver": "^6.0.0",
     "stats-js": "^1.0.0",
-    "three": "~0.104.0",
-    "uWebSockets.js": "uNetworking/uWebSockets.js#v15.10.0",
-    "uws": "^9.14.0"
+    "three": "~0.105.0",
```
Qaomen: Updating this causes WebGL warnings to appear.

Qaomen (changes requested):
I don't know how the webGL warnings were removed earlier, but they seem to be back with the new Three.js version. Do you think you could fix those (if you remember how you did it last time)?
Also, `reconnecting-websocket` still seems to be outdated, probably since it's missing a `~` before the version. We should probably update this too.

Qaomen:
On a side note, `ammo.js` is still outdated (according to the lockfile, it uses v0.0.2), but it seems to change the way it initialises. I'll take a look at it myself sometime later.

Qaomen (approved):

PR #236, opened 2019-06-20
Marble rotation fix, minor restructures

### LINKED ISSUE #212: Marble rotation data optimisation
**Description**
Marble rotations are sent and processed as quaternions. To save on bandwidth, we could convert them to Euler angles with implicit XYZ rotation order and convert them back on the client. This reduces raw rotation data by 25% (4 numbers per rotation vs 3).
Since converting quaternions to/from Euler angles does require some math, we should check whether it notably affects server/client performance with large marble counts just in case, and reconsider if it's an issue.

- [ ] Convert rotations to Euler angles with XYZ rotation order before sending them to the client.
- [ ] Client should convert rotations back to quaternions upon receiving them.
- [ ] Performance impact should be checked for large marble count.

Qaomen:
**Changes**
- Server now syncs angular velocity rather that quaternions (also reducing total physics data size by a some amount).
- MarbleMesh now has a "main" Three.js object with the mesh and name tag as child objects, so we can modify the mesh's rotation without the name tag spinning.
- Name tags use layers for visibility instead.
- Removed MarbleManager's `setMarbleTransforms` since it was a very specific use case. Transforms can be set directly instead.

Angular velocity isn't interpolated since the difference would hardly be noticeable (if at all), and not "correct" in either case. Maybe I'm cutting corners a little, but I'd rather not let the code do that math & array copying unless it's really worth it.

**Referencing issues**
Closes #212

### DIFF
--- a/src/client/marble-manager.js
+++ b/src/client/marble-manager.js
@@ -7,21 +7,17 @@ let _userData = Cookies.getJSON("user_data");
 
 // This module manages all the marbles that physically exist in the scene.
 let marbleManager = function() {
-	let _marbles = [], // Array of marbles that currently exist in the scene
-		_skins = {},
+	let _skins = {},
 		_fallbackSkin = null;
 
 	return {
+		marbles: [], // Array of marbles that currently exist in the scene
 		marbleGroup: null, // Group containing marble instances
-		marbleNamesGroup: null, // Group containing name sprites
 		marbleGeometry: null,
 
 		initialize: function() {
 			this.marbleGroup = new THREE.Group();
-			this.marbleNamesGroup = new THREE.Group();
-			this.marbleNamesGroup.renderOrder = 1;
 			renderCore.mainScene.add(this.marbleGroup);
-			renderCore.mainScene.add(this.marbleNamesGroup);
 
 			// Default marble model
 			this.marbleGeometry = new THREE.SphereBufferGeometry(1, 32, 32);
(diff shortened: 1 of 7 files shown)


src/client/marble-manager.js
```diff
@@ -141,14 +115,18 @@ const MarbleMesh = function(marbleData) {
 	this.mesh.receiveShadow = config.graphics.receiveShadow.marbles;
 
 	// Highlight own name
-	let nameSpriteOptions = {};
+	let nameSpriteOptions = {
+		renderOrder: 1
```
Zemanzo: Move this to the `makeTextSprite` function

src/client/marble-manager.js
```diff
@@ -178,6 +156,8 @@ const makeTextSprite = function(message, options = {}) {
 
 	let sprite = new THREE.Sprite(spriteMaterial);
 	sprite.scale.set(0.3, 0.1, 1.0);
+	sprite.layers.disable(0);
+	sprite.layers.enable(renderCore.SPRITE_LAYER);
 	if (options.renderOrder) sprite.renderOrder = options.renderOrder;
```
Zemanzo: If this is set by default, you can get rid of the check

Zemanzo (changes requested):
Good stuff! And I certainly do not think you're cutting corners here. It looks good, no need to pointlessly ruin performance.

Just a minor thing with the nameSprite render order, but you already knew that ;)

src/client/marble-manager.js
Zemanzo: nvm I am a dumb

Zemanzo (approved):

Issue #233, opened 2019-06-20
Dev bot commands


Qaomen:
**Description**
Once the server is running, there is no easy way to influence configuration options or alter the state of the game (e.g. resetting a race). Some dev-only bot commands could be useful here!
Player commands would be nice to have in the future as well, but these won't be part of this issue.

**Minimum Acceptance Criteria**
- [x] A command to reset a race (similar to `!end` but without considering the race finished w/ a database entry & points).
- [ ] Commands to kick and/or ban a user.
- [ ] Commands to alter marble rules (max marbles, enter/dnf timeouts, auto-joined bot count etc).
- [ ] The bot should display a message for commands involving configurable values, to confirm and/or read the current value.

Zemanzo:
Useful guide for setting up something proper: https://discordjs.guide/command-handling/#individual-command-files

PR #237, opened 2019-06-21
Banned users can no longer enter marbles or chat

### LINKED ISSUE #166: Banned users
**Description**
Users that are banned from the server should not be able to participate in marble races or send messages to the discord channel through the client.

- [x] Banned users can't participate in races.
- [x] Banned users can't send messages in the game channel.
- [x] Banned users should be flagged as such in the database, in case those leaderboard entries are chosen to be hidden in the future.

Zemanzo:
That said, banning users can be tricky. If they only enter through the embed and are not in the server itself, it might be impossible to get their ID (for mods, at least) and thus being impossible to ban anyway. This could be solved by adding a bot command that fetches ID by user name (something for #233?)

Closes #166

### DIFF
--- a/src/server/chat/discord-manager.js
+++ b/src/server/chat/discord-manager.js
@@ -51,10 +51,12 @@ const discordManager = function() {
 			this.client.on("error", console.error, console.error);
 
 			this.client.on("guildBanAdd", function(guild, user) {
+				log.info(`DISCORD: ${"Banned user".red} ${user.username}#${user.discriminator} (${user.id})`);
 				db.user.setBanState(true, user.id);
 			}, console.error);
 
 			this.client.on("guildBanRemove", function(guild, user) {
+				log.info(`DISCORD: ${"Unbanned user".green} ${user.username}#${user.discriminator} (${user.id})`);
 				db.user.setBanState(false, user.id);
 			}, console.error);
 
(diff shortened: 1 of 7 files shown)


src/server/database/initializer.js
```diff
@@ -15,6 +15,7 @@ module.exports = function(db) {
 				timestamp_refresh_last INTEGER,
 				time_refresh_expire INTEGER,
 				scope TEXT,
+				is_banned INTEGER,
```
Qaomen, line 18: This flag doesn't have a default value set in `_insertNewUserEmbed` and `_insertNewUserDiscord`.

src/server/chat/discord-manager.js
```diff
@@ -49,6 +49,16 @@ const discordManager = function() {
 			}, console.error);
 
 			this.client.on("error", console.error, console.error);
+
+			this.client.on("guildBanAdd", function(guild, user) {
+				db.user.setBanState(true, user.id);
```
Qaomen, line 55: I think it would be nice to log these ban/unban events for the server. Wouldn't hurt to be able to read them back, at least?

Qaomen (changes requested):
Banning is a trickier thing than I thought, like you mentioned.
I'd like to see that users are notified in the client if they are banned, since there's no feedback given at the moment. But since it wasn't part of the acceptance criteria, I'll leave it to you whether to address that now or later. We'll likely have to take a closer look at moderating in general, anyway.

src/server/database/initializer.js
Zemanzo: Not really necessary since it only checks for `is_banned != 1` (in which case both `0` and `null` would return true)
Any arguments on why it should have a default value?

src/server/chat/discord-manager.js
Zemanzo: Discord has an audit log which you can also check. Regardless, I'll still add them as it is a bit nicer for debugging in case anything goes wrong.

Zemanzo:
> I'd like to see that users are notified in the client if they are banned

@Qaomen Heck, I thought about it during testing but then forgot again! I'll add it.

src/server/database/initializer.js
Qaomen: Because the value is assumed to be an integer (`0` in this case). With the current check this isn't a problem, but it could be in the future if our code changes. If we checked `is_banned == 0` instead, we get either false or true depending on whether `is_banned` is `null` or `0`, and both are currently possible.
If we initialise the value, we can safely assume it's the type we want it to be.

Zemanzo: Fair enough!

Qaomen (approved):

PR #246, opened 2019-07-08
Minor visual chat improvements

Sanquinary:
**Title**
Added word-break, less transparency of chat window and fixed better handling of it

**Description**
Transparency is now configured in it's own class element `semiTransparent` and not directly in `client/chat/chat.js`
Messages now gets broken anywhere if it exceeds chat width. (previously it only broke if character had space after)

### DIFF
--- a/public/styles/chat.css
+++ b/public/styles/chat.css
@@ -94,7 +94,6 @@ header {
 	.message {
 		display: flex;
 		align-items: baseline;
-		word-break: break-word;
 		max-width: 100%;
 		overflow: hidden;
 		padding: 0 8px;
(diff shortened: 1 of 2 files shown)


public/styles/chat.css
```diff
@@ -90,6 +94,7 @@ header {
 	.message {
 		display: flex;
 		align-items: baseline;
+		word-break: break-word;
```
Zemanzo: This results in inconsistencies when it comes to cross-browser support. In Firefox, the username will be above the message, whereas in Chrome the username is to the left of the message. I would suggest the following as a fix:

```css
.content {
  white-space: nowrap;
}
```

```css
.text {
  word-break: break-word;
  white-space: normal;
}

Zemanzo (changes requested):

Zemanzo (approved):

PR #247, opened 2019-07-10
Socket refactor

### LINKED ISSUE #200: Socket refactor
**Description**
The current socket implementation could do with cleanup, for the server and possible the client to some degree. This may slightly improve performance on the server as well.

- [ ] Code should be cleaned up and easy to read.
- [ ] Socket code should conform to coding standards.

Zemanzo:
- Removed unused code
- Renamed files
- Split gameplay socket from chat socket, this avoids a circular dependency with `game.js` and `messages.js`
- Adjusted code to conform to standards

This PR does not modify anything on the client side.

**Referencing issues**
Closes #200


src/server/chat/messages.js
```diff
@@ -1,64 +1,70 @@
...
+					id,
+					username,
+					_createAttributesObject(messageContent)
+				);
+			}
 
-	else if (messageContent.startsWith("!end") && isDeveloper) {
-		game.end();
-	}
+			// -- Developer commands
```
Qaomen, line 49: If the dev-only commands start here, it might look a little better to nest all dev commands in one `isDeveloper` check.

src/server/network/socket-gameplay.js
```diff
@@ -0,0 +1,55 @@
+const socketsHelper = require("./sockets-helper");
+const log = require("../../log");
+const db = require("../database/manager");
+
+const socketGameplay = function(getInitialDataPayload) {
```
Qaomen, line 5: This currently isn't a module, but a constructor of some kind (in`game.js` it's called as a function to create an instance).
It's currently part of a circular dependency (hotfixed by passing a function as parameter), and could be resolved by changing the architecture: Network-code is to handle tickrate & when to emit messages (requesting the data from `game`), removing the need for the `game` module to refer to and/or manage any kind of networking.

Edit: It may actually stay instantiate-able (probably not a word but oh well) but leaving the full message here just in case.

src/server/game.js
```diff
@@ -227,11 +227,20 @@ let game = function() {
 		}
 	};
 
+	let getInitialDataPayload = function() {
```
Qaomen, line 230: Moving this function should be reverted once the game-network circular dependency is resolved.

src/server/network/socket-gameplay.js
```diff
@@ -0,0 +1,55 @@
...
+	return new socketsHelper.Socket(
+		// Route
+		"/gameplay",
+
+		// Options
+		{
+			compression: 0,
```
Qaomen, line 12: There seems to be no issues if we enable compression, so I'd say it worth to turn it on?

src/server/network/sockets-helper.js
```diff
@@ -0,0 +1,136 @@
...
+			log.warn(`µWS: Failed to listen to port ${config.uwebsockets.port}`.cyan);
+		}
+	});
+
+	return {
+		stopListening: function() {
+			uWS.us_listen_socket_close(_token);
+		},
+
+		Socket: function(
```
Qaomen, line 31: Given that this creates a `Socket` instance and that the chat/gameplay sockets override variables and functions, it may look a lot cleaner if we use proper inheritance for this instead. We only have those two socket types after all.

src/server/network/sockets-helper.js
```diff
@@ -0,0 +1,136 @@
...
+			// Additional functionality when connection to socket is made (open)
+			let _open = options._open = options.open;
+			delete options.open; // avoid conflict with Object.assign;
+
+			// Additional functionality when connection to socket is lost (close, disconnect)
+			let _close = options._close = options.close;
+			delete options.close; // avoid conflict with Object.assign;
+
+			// Create list that contains all open socket connections
+			this._list = [];
```
Qaomen, line 64: Could be just a little more descriptive than a list called `_list` :D

Qaomen (changes requested):
Overall, it looks nicer and cleaned up. Wooh!
Though it looks like it could do with some architectural changes to fix circular dependencies and have a clearer control flow. I'll probably tackle these myself but input/approval on the issues is still very much appreciated!

src/server/network/socket-gameplay.js
Zemanzo: Since we already compress it on our end, this will only slow things down. Unless you observed some benefits?

Qaomen: We currently don't compress gameplay data with `pako`, we just pack it together with MessagePack and leave it at that! It could do with further testing, but I'm making the assumption that `pako` may not be optimised for performance as much as network-specific compression.

Zemanzo: Ah, I thought we did! Then it might be a good thing to look into.

Zemanzo:
Closed by #270

PR #249, opened 2019-08-16
Skins refactor (Part 1)

### LINKED ISSUE #227: Skins refactor
**Description**
The current implementation of marble skins is functional, but not very flexible. An improved design can be found in the TDD.

- [x] Have a single JSON file that serves as an entry point for skins.
- [x] Have the ability to add a normal map and adjust other material properties such as roughness and metalness.
- [x] Add a premium-only flag for skins.
- [x] Add the ability to have several keywords for the same skin.
- [x] Add a flag for skins that can utilize custom colors.
- [x] Make sure the skins preview mode uses shared code for generating marbles (see comment on #223)

Zemanzo:
I almost pulled another #149 but then thought about the horrors, and decided to cut the feature here for now. Don't be scared of the amount of files / lines, a bunch is just skin meta files or custom shaders.

Things that are NOT in this PR:
- Actual premium skin functionality. It's an unused value right now. Anyone could use premium skins.
- "Thumbnails" for skins. The little icons next to the name in the list on the left. Those are not generated for custom shader skins, and some regular skins could use improved versions as well.
- All vertex shaders are equal at this time, and most custom shader skins will probably use the same vertex shader. I will make it so the basic vertex shader is the default, and only has to be downloaded once. (Custom ones can still be used, if specified in the meta.json file)

I will make issues for all of these.

**Referencing issues**
Closes #227


src/server/chat/messages.js
```diff
@@ -15,7 +15,7 @@ const createAttributesObject = function(messageContent) {
 		}
 
 		if (!skinId) {
-			skinId = skins.idList.includes(messageSections[i]) ? messageSections[i] : undefined;
+			skinId = skins.idList[messageSections[i] ? messageSections[i] : undefined];
```
Qaomen: Not sure what the logic is here? Currently it checks whether `messageSections[i]` resolves to `true`, not whether it exists in the list.

src/client/render/render-core.js
```diff
@@ -83,7 +89,7 @@ let renderCore = function() {
 			} else { // Initialize
 				this.mainScene = new THREE.Scene();
 				_renderer = new THREE.WebGLRenderer();
-				_renderer.debug.checkShaderErrors = false;
+				_renderer.debug.checkShaderErrors = true;
```
Qaomen: Sort of indirect leftover logging, I think?

src/client/marble-skins.js
```diff
@@ -0,0 +1,175 @@
...
+					})
+					.catch(() => {
+						// Skin failed to load, so remove it from the skins list
+						delete _skins[skinId];
+					});
+			} else {
+				// Skin meta has already been retrieved before, so we only have to return it. Since the
+				// return value is always expected to be a promise, we will return a promise that
+				// immediately resolves to the already loaded material.
+				return Promise.resolve(_skins[skinId].material);
```
Qaomen: Because the creation of the material also sets the colour, any marbles that spawn with the same skin have the colour of the first marble that spawned with it.
If you spawn multiple marbles at once (or refresh the client) the colours appear to be fixed, but that's because all the promises are being resolved at the same time, and those skins aren't re-used at that point (meaning we're loading the same skin multiple times when that happens).

We may have to store the loaded shaders/textures per skin instead, and then generate materials from them.

src/client/render/render-core.js
```diff
@@ -29,6 +29,11 @@ let renderCore = function() {
 		// Make updates
 		renderCore.updateCallback(deltaTime);
 
+		// Update shaders
+		for (let shaderMaterial of renderCore.shaderMaterials) {
+			shaderMaterial.uniforms["time"].value += deltaTime;
```
Qaomen: Uniforms are values that are shared on the GPU, meaning you'll only have to do this once for all shaders to be updated!
(You can use the skin loading bug to see this: If you spawn 50 fancy marbles at the same time, the shader is loaded + added to this list 50 times and progresses way too fast. Then if you add a different fancy marble, it matches the speed of the other shader.)

Also, it might be worth it to reset this to 0 between rounds, in case people leave their clients open a long time and start to get rounding errors. Don't know how fast that would happen, though.

Qaomen (changes requested):
Got some juicy bugs for you to fix.
Love how the skins are managed server-side btw, great stuff! :D

src/server/chat/messages.js
Qaomen: Just to add: It works, but it looks accidental?

src/client/render/render-core.js
Zemanzo: Ye oops

src/client/render/render-core.js
Zemanzo: This makes a lot of sense with the issues I've been having. I had a bit of trouble getting the uniforms to work at all, so once it finally did the thing I was too scared to change it. Good to know!

src/client/marble-skins.js
Zemanzo: Ah yes, the custom color thing was pretty much added in the last, as an afterthought. I kind of want to re-do it so that we're not stuck with the default blend mode for the rest of the textures.

Regarding how to store the skins, we might want to do that only for the skins that have custom color enabled.

I'll think about this some more.

src/server/chat/messages.js
Zemanzo: I'm pretty sure this was brilliant when I wrote it, but I agree it looks kind of confusing. Not sure if I should change it, any suggestions? I could also just add some comment to clarify it.

Zemanzo: How about `skinId = messageSections[i] ? skins.idList[messageSections[i]] : undefined;`?

src/client/marble-skins.js
Qaomen: The colour issue is resolved, but it's technically still bugged in the way that it will load the same skin multiple times if multiple marbles spawn in at once.

In the end they all load the same thing and overwrite each other in `_skins[skinId]` meaning there's no lasting issue, but letting the client load the same skin multiple times doesn't look very good still. Perhaps you can resolve it by storing the skins as Promises with the actual skin Object as it's resolved return value.

src/server/chat/messages.js
Qaomen: Checked again just in case and `messageSections[i]` will always be a string, so we don't need to check whether it's undefined.

So technically, just `skinId = skins.idList[messageSections[i]];` will suffice! Non-existing skins will return `undefined`. But since that might be a bit confusing, it could probably do with a comment in that case.

src/client/render/render-core.js
```diff
@@ -61,6 +64,9 @@ let renderCore = function() {
 		activeCamera: null,
 		freeCamera: null,
 		trackingCamera: null,
+		shaderUniforms: {
+			"time": { value: 1.0 }
```
Qaomen: I missed this last time, but since this value represents the time elapsed, it should probably start at 0. This really doesn't change anything but eh.

Qaomen (changes requested):

Qaomen (approved):

PR #252, opened 2019-09-03
Tracking camera zoom control & improvements

Zemanzo:
So this wasn't an issue, but I worked on it anyway. With this feature, you can scroll the mouse wheel to zoom the tracking camera.

This somehow also made it very tempting to look around the marble, so it might be worth considering a feature that would allow the user to control the angle.


Zemanzo:
Don't merge this yet, there's an issue that will alter the zoom level even when scrolling outside of the viewport.

Zemanzo:
Should be good for a review / merge now!

Qaomen (approved):

PR #256, opened 2019-09-03
Update dependencies, imports & webpack config

### LINKED ISSUE #2: Importing three.js examples as modules
Three.js examples provide excellent snippets for loaders, camera configurations and more, but aren't modular (they rely on the window.THREE global) and thus cannot easily be imported. This is not directly an issue for this project, however, needs to be worked around. You can find full details on the three.js issue [here](https://github.com/mrdoob/three.js/issues/9562).

~~The current workaround is done by using [this plugin for webpack](https://github.com/wildpeaks/package-three-webpack-plugin).~~

Zemanzo:
This solves some vulnerabilities, but more excitingly, allows us to import three.js examples as modules! That removes the need for some nasty webpack modification. It also allows for proper tree shaking, where we can ignore all unused three.js code. Should give some significant loading benefits! I'll save that for another PR though, might be a bit of a hassle to do. Take a look at `custom-materials.js` in this PR for an example.

**Referencing issues**
Closes #2 at long last...


Qaomen (approved):
Updated some additional stuff etc. Just merge whenever it's all good on your side!

PR #263, opened 2019-11-21
Improved hud notifications & Altered client header

### LINKED ISSUE #250: HUD Notifications block camera buttons
**Expected behavior**
HUD notifications will keep to their darn selves!

**Actual behavior**
The left and right sides outside of the HUD notifications will block overlay buttons by overlapping them, even though it doesn't appear to block them. Only the camera style buttons are affected due to their positioning, and the issue disappears once the HUD notifications are gone.

**Reproduction steps**
1. Load the client page.
2. Wait for HUD notifications appear (e.g. the loading/loaded notifs).
3. Try to click the camera buttons while the notification is visible.

**Additional context**
There's probably a more graceful and desirable way to fix this other than re-ordering the overlay elements, so I prefer to leave it to someone who knows said pretty and clean fix (unless they're fine with the re-arranging stuff).

Zemanzo:
Notifications:
- Added progress indicator functionality to notifications.
- Ability to change attributes after notification construction.

Level loading emits events to indicate loading progress, which are used in the level loading notification.

Client header has more visible and stylistically attractive links. It also includes a link to the Discord server now.

**Referencing issues**
Closes #250


templates/client.mustache
```diff
@@ -30,8 +30,20 @@
 				{{#gitHash}}<div id="releaseHash">{{gitHash}}</div>{{/gitHash}}
 			</div>
 			<nav id="externalLinks">
-				<a href="https://github.com/Zemanzo/web-marbles" id="github" title="Contribute to the source!"><i class="icon-github-circled"></i></a>
-				<a href="https://www.patreon.com/webmarbles" id="patreon" title="Support us on patreon!"><img src="images/patreon.svg" alt="Patreon"></a>
+				{{#invitelink}}
+				<a href="{{invitelink}}" id="discord">
+					<img src="images/wumpus_white.svg" alt="Wumpus">
+					<span>Join the Discord!</span>
```
Sanquinary, line 36: While I do like the new buttons, I preferred the minimalistic look mainly because it wasn't distracting for the gameplay. In my opinion, the buttons should not be in focus that much and perhaps the `<span>` tags for the new buttons should only pop in when hovered?

public/styles/client.css
```diff
@@ -356,52 +358,66 @@ body {
...
+				#notifications .notification-progress-step.pending {
+					background-color: #0009;
+				}
+
+				@keyframes notification-progress-step-loading {
+					0% { background-color: #f90f; }
+					100% { background-color: #f903; }
+				}
+
+				#notifications .notification-progress-step.loading {
```
Sanquinary, line 410: Great addition with the `loading` step, I think it can be further improved if a class to the `notification-progress-step` for a `.failed` state is added.

Reason being is that it might be confusing for the user when the `loading` icon is still showing while an error has occurred. Gives an indication of that something is happening in the background while it really isn't.

src/client/level-loader.worker.js
```diff
@@ -5,9 +5,11 @@ onmessage = function(message) {
...
 
 let loadLevelData = function(url) {
+	postMessage({ event: "downloadStart" });
 	fetch(`${url}`)
 		.then((response) => {
 			if(response.ok) {
+				postMessage({ event: "downloadComplete" });
 				return response.arrayBuffer();
 			} else {
 				loadingFinished(false, `Unable to retrieve ${url} from server: ${response.status} - ${response.statusText}`);
```
Sanquinary, line 15: On Mac OSX 10.15.2, Safari 13.0.4 - running this branch locally failed to render the scene and was stuck 'Downloading...' the scene. This error was never caught and I was left with the notification displaying for an infinite amount of time which is something I wouldn't consider the best UX.

However, this issue might be very browser-specific or caused by other factors and might be expected. It's worth a note nonetheless.

Sanquinary (commented):
Nice work, I really like the new looks of it and the end-user gets a really nice overview of what's going on behind the scenes. The displayed notification text makes sense and I think there's no changes needed to be made there.

src/client/level-loader.worker.js
Zemanzo: Since we haven't aimed to support IOS, OSX and Safari, I'm going to let this one slip by for now. If it still occurs in `dev` or `master` after merging, please make it an issue so it does not get forgotten.

Sanquinary (approved):
Great job, looks amazing! 🥇 👍

PR #265, opened 2019-11-27
v1.3.1

Zemanzo:
Changes since v1.3.0:
- Level loading notification now indicates progress and no longer disappears before the level is loaded.
- Outgoing links in client header have their style altered, and now also includes a link to the Discord server.


Zemanzo:
~~Wait with reviewing / merging, I have another fix that needs to be in this.~~

Zemanzo:
Ok we good now.

Qaomen (approved):

PR #270, opened 2020-03-11
Socket refactor, take 2

### LINKED ISSUE #200: Socket refactor

Qaomen:
Since the original branch was *very* out-of-date, here's a new branch with the same initial changes but with some refactoring on top of that:
- Cleaned up the `Socket` object. The main difference is that it uses an `EventEmitter` to signal events (`open`, `close`, and `message`), rather than overriding existing functions.
  - The `gameplay/chat` specialisations have been removed in favour of event callbacks. This means we also no longer have that circular dependency problem!
  - The `Socket` instances for the chat and gameplay are now managed by the `game` and `DiscordManager` modules.
- Cleaned up `DiscordManager`, mainly regarding private variables and the `db` parameter.
- The user auth check is now part of the chat socket logic, rather than the gameplay one.
- Added init/stop function for `game` to make the relevant code a bit cleaner, though `game` still needs work as a whole.

**Referencing issues**
Closes #200


src/server/chat/discord-manager.js
```diff
@@ -1,47 +1,121 @@
...
+			_chatClientSocket = new socketsHelper.Socket("/chat", {
+				compression: 1,
+				maxPayloadLength: 128 * 1024,
+				idleTimeout: 3600
+			});
 
-			this.client.on("ready", () => {
+			_chatClientSocket.eventEmitter.on("open", _onClientChatConnected);
+			_chatClientSocket.eventEmitter.on("close", _onClientChatDisconnected);
+			_chatClientSocket.eventEmitter.on("message", _onClientChatMessage);
```
Zemanzo: The naming here isn't very consistent. I'd prefer to see the private functions also start with `_onChatClient`, because starting with `_onClient` made me confuse it with the Discord client for a moment.

Zemanzo (changes requested):
Just a minor naming thing, lovely stuff overall!

I am tempted myself to tweak `discord-manager.js` some more, by moving all `_discordClient` event functions to private functions and cleaning up the authorize and refresh client functions. I'd rather tackle that in a separate PR though, so we can get these changes through.

Zemanzo (approved):

PR #273, opened 2020-04-10
Update dependencies

Zemanzo:
Did the usual round of updating dependencies.
Skipped on discord.js for now, as it is a major release and will probably break some stuff. It might be nice to look into updating it later because it does have some improvements under the hood.

From what I've seen and tested, these new dependency versions shouldn't break anything.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
   "dependencies": {
     "@discordjs/uws": "10.149.0",
     "ammo.js": "kripken/ammo.js#ac08f42d3995cfa2b719aa8a9bb5b766c7f63404",
-    "better-sqlite3": "^5.0.1",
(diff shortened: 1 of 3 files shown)


Qaomen (changes requested):
We're getting Three.js warnings on the client, seems that they've moved some functionality around:
```
THREE.WebGLRenderer: .gammaInput has been removed. Set the encoding for textures via Texture.encoding instead.
THREE.WebGLRenderer: .gammaOutput has been removed. Set WebGLRenderer.outputEncoding instead.
```
The rendering doesn't look messed up (though I haven't done a side-by-side comparison either), but since we've explicitly set these it's probably best to fix them.

Qaomen (approved):

PR #279, opened 2020-04-22
Server-side game.js refactor, dynamic level loading

### LINKED ISSUE #53: Server crashes due to unhandled errors
**Expected behavior**
The server should handle potential errors gracefully and not crash.

**Actual behavior**
The server doesn't explicitly handle some errors that could occur, causing it to crash.
For instance, it crashes if it loses connection with the internet (1), or crashes due to a discord authorisation error (2). Possibly, there are more scenarios.

**Reproduction steps**
(1) Start the web server and disable your internet connection.
When a web server is running and it temporarily loses connection, the following crash occurs:
```
Error [ERR_UNHANDLED_ERROR]: Unhandled error. ([object Object])
    at Client.emit (events.js:178:17)
    at WebSocketConnection.onError (D:\Repos\Git\web-marbles\node_modules\discord.js\src\client\websocket\WebSocketConnection.js:374:17)
    at WebSocket.onError (D:\Repos\Git\web-marbles\node_modules\ws\lib\event-target.js:128:16)
    at WebSocket.emit (events.js:189:13)
    at _receiver.cleanup (D:\Repos\Git\web-marbles\node_modules\ws\lib\websocket.js:211:14)
    at Receiver.cleanup (D:\Repos\Git\web-marbles\node_modules\ws\lib\receiver.js:557:13)
    at WebSocket.finalize (D:\Repos\Git\web-marbles\node_modules\ws\lib\websocket.js:206:20)
    at TLSSocket.emit (events.js:189:13)
    at emitErrorNT (internal/streams/destroy.js:82:8)
    at emitErrorAndCloseNT (internal/streams/destroy.js:50:3)
Assertion failed: err == 0, file src\uv-common.c, line 686
error Command failed with exit code 3221226505.
```

(2) Unclear cause, but has to do with requesting an access token.
```
Error [ERR_HTTP_HEADERS_SENT]: Cannot set headers after they are sent to the client
    at ServerResponse.setHeader (_http_outgoing.js:470:11)
    at ServerResponse.header (D:\Repos\Git\web-marbles\node_modules\express\lib\response.js:767:10)
    at ServerResponse.send (D:\Repos\Git\web-marbles\node_modules\express\lib\response.js:170:12)
    at ServerResponse.json (D:\Repos\Git\web-marbles\node_modules\express\lib\response.js:267:15)
    at ServerResponse.send (D:\Repos\Git\web-marbles\node_modules\express\lib\response.js:158:21)
    at request.post.then (D:\Repos\Git\web-marbles\src\server\chat\discord-manager.js:157:11)
```

### LINKED ISSUE #137: Game loop
**Description**
The server should have a completed game loop, which includes level rotation and/or voting (depending on the final design). Not everything needs to be finalized, as long as the main flow is there.

- [ ] Design should be clearly documented.
- [ ] Relevant game states should be added (e.g. round end for result display, level changes, etc).
- [ ] Levels should be dynamically loadable.
- [ ] Levels should be validated before use.

### LINKED ISSUE #173: Level rotation functionality
**Description**
We should have a base functionality to support level changing in between rounds. This does not yet include community voting.

- [ ] During server initialisation, the server should check for and keep a list of available and valid levels.
- [ ] During the "finished" or "waiting" period, an admin should be able to change the level with a command.

#137

Qaomen:
Whoops I accidentally finished dynamic level loading for all the levels we don't have!

The main changes in a nutshell, or two, or four:
- Game state changes are now entirely handled in `_setCurrentGameState` and the `_onStateX` functions.
- LevelManager is now easier to use (no properties wrapped in a `Promise`), and validates all levels (for both server and client) before the game is initialised.
  - Validation can be disabled, in case it slows down development/testing due to many large level files.
- The level can now be changed with `!level` by devs, before marbles are entered.
- Added a `STATE_LOADING` game state for level changes.
- Added `!start` to skip the countdown (also dev only).
- Slightly reduced the clutter in `web-marbles.js`.
- Fixed marble skins not actually checking `allowCustomColor`.
- Fixed a bug where ending a race and discarding the results could cause an invalid state for clients when they refresh.

**Referencing issues**
Closes #53 (I couldn't reproduce this. Since it's an outdated issue too, I think we can consider it fixed)
Closes #137 
Closes #173


src/server/chat/commands.js
```diff
@@ -54,6 +54,14 @@ const commands = function() {
 
 
 		// Developer commands
+		{
+			// Starts the race early, in case you gotta go fast but the countdown is being a bully
```
Zemanzo, line 58: Huh, this makes me think we should have a description field, so we can have a `!commands` command of some sort, listing all available commands with their descriptions. Something for another PR of course, but I figured I would write it down as to not forget it.

src/server/game.js
```diff
@@ -255,13 +237,160 @@ let game = function() {
...
+		case gameConstants.STATE_FINISHED:
+			stateName = "STATE_FINISHED";
+			_onStateFinished();
+			break;
+		default:
+			stateName = "Unknown";
+		}
+
+		// Print the state change only once (e.g. aborted races technically "finish" and go to the waiting state right away)
+		if(newState == _currentGameState)
```
Zemanzo: When possible, it is preferred to use a triple equals sign `===`

src/server/game.js
```diff
@@ -255,13 +237,160 @@ let game = function() {
...
 
+	// Sets currentGameState and informs all connected clients about the state change
+	let _setCurrentGameState = function(newState) {
+		_currentGameState = newState;
+		_netGameUpdate.g = newState;
+		_triggerNetworkUpdate();
+		clearTimeout(_gameStateTimeout); // Always clear any remaining timeouts if we switch states
+
+		let stateName;
+		switch(newState) {
```
Zemanzo, line 197: So I really do not like this switch statement. It's just so heckin' repetitive. My solution would be pretty significant but I do feel it would be worth it. I suggest doing the following:

Change `game-constants.json` to a .js file that exports every set of constants individually such as:
```js
// We use module.exports because it's used on both the client side and the server side.
module.exports.GAME_STATES = {
	"WAITING": 0,
	"ENTER": 1,
	"STARTING": 2,
	"STARTED": 3,
	"FINISHED": 4,
	"LOADING": 5,
}
module.exports.RECORDS = {
	"NONE": 0,
	"PB": 1,
  "WR": 2
}
```

Then in `game.js` do the following:
```js
import GAME_STATES from "../game-constants";

...

_setCurrentGameState = function(newState) {
  Object.keys(GAME_STATES).forEach(state => {
    if (newState === GAME_STATES[state]) {
      stateName = state;
      const functionName = `onState${state.substring(0, 1).toUpperCase() + state.substring(1)}`;
      this[functionName]();
    }
  }
}
```

Alright yeah that might not be the cleanest solution either. There must be something better though... Let's at least discuss it.

src/server/game.js
```diff
@@ -9,8 +9,6 @@ const permissions = require("./chat/permissions");
 const skins = require("./skins");
 const msgPack = require("msgpack-lite");
 
-const anyColorAllowed = ["abstract", "marble", "default", "swirly", "squares"]; // Will (probably) be changed to a meta file later on, next to the skin file?
-
 function Marble(id, entryId, name, attributes = {}) {
```
Zemanzo: Should we move this class to its own file? (Since I started doing programming professionally now, I prefer every class in their own file)

src/server/game.js
```diff
@@ -618,26 +645,6 @@ let game = function() {
...
-// The actual getTimeLeft function
-function _getTimeout(id) {
-	let m = _timeoutMap[id]; // Find the timeout in map
-
-	// If there was no timeout with that id, return NaN, otherwise, return the time left clamped to 0
-	return m ? Math.max(m[1] + m[0] - Date.now(), 0) : NaN;
-}
 
 // Generates a random color in the HEX format
 function _randomHexColor() {
```
Zemanzo: Perhaps we can also move utility functions like these to their own file.

src/server/game.js
```diff
@@ -357,30 +464,19 @@ let game = function() {
...
-				this.currentGameState === gameConstants.STATE_FINISHED
-				// Check whether we have reached the maximum marble limit
-				|| _marbles.length >= config.marbles.rules.maxMarbleCount) return;
+			if (_currentGameState === gameConstants.STATE_FINISHED || _currentGameState === gameConstants.STATE_LOADING || _marbleLimitReached)
+				return;
 
 			// Start physics simulation if this is the first marble
-			if(_marbles.length === 0) physics.world.startUpdateInterval();
+			if(_marbles.length === 0)
+				physics.world.startUpdateInterval();
```
Zemanzo: This does not adhere to our style guide. Either keep it on a single line, or add brackets.

src/server/game.js
```diff
@@ -396,140 +492,74 @@ let game = function() {
...
 
-				return true;
-			} else {
-				return false;
-			}
+			levelManager.loadLevel(levelName).then( (levelData) => {
+				_currentLevel = levelData;
+				if(!levelData) {
+					// Well this is awkward
+					// This really shouldn't happen if all levels were validated earlier, so we're probably in booboo town already
```
Zemanzo, line 557: :'D

Zemanzo (changes requested):
That's a PR alright. Nice job with both the refactor and the level loading! Here's a couple more things to ponder over, looking forward to merging this soon!

src/server/game.js
Qaomen: I can't remember why we actually have the single-line rule, but I'll fix it either way!

src/server/chat/commands.js
Qaomen: To add to this, it may also be worth it to let the bot respond to some of the commands, e.g. `!level` without a parameter would post the name of the level in the chat, or a non-dev who tries `!lotsofbots` could get a message stating the command is not allowed.
But yeah, something for another PR!

src/server/game.js
Qaomen: (Good on ya!)
This should be possible, though looking at it I'll need to rework the way marbles work with the game and physics, since there's a circular dependency related to the `onFinished` callback.
This should ultimately decouple physics and the marbles/game a bit further though, and we'll need that if we ever want to start running the physics on a different thread!

src/server/game.js
Zemanzo: I think the main reason is that a second line without brackets could be confusing. The only difference that makes it visible that it's a conditional statement is the indent. Having it on a single line helps, considering you read left to right.
However, nowadays I'm tempted to say to change the code style to always include brackets. It might become overly verbose in some locations but overall it should help with readability.
Then again, why bother to keep changing things, might as well just stick with what we have, it works fine.

Zemanzo (approved):

Issue #199, opened 2019-06-11
Audio settings


Qaomen:
**Description**
We should have a volume control for sound effects. This may be combined with the UI for the radio feature (#158), or with the graphics settings UI (#155).

**Minimum Acceptance Criteria**
- [ ] Clients should be able to influence the sound effect volume through the use of a slider.
- [ ] The volume setting should be saved in between sessions.

Issue #197, opened 2019-06-10
Add total amount of entrants on race leaderboards


Zemanzo:
**Description**
Currently, there is no way to see how many entrants there were in the race unless you scroll all the way to the bottom. Having a number at the top of the leaderboards UI makes it much easier to see.

**Minimum Acceptance Criteria**
- [ ] Display total amount of entrants the race had at the top of the leaderboards at the end of the race.

Issue #158, opened 2019-05-10
Radio


Zemanzo:
**Description**
Add an interface for playing music, likely a radio stream ( https://plaza.one/ )

**Minimum Acceptance Criteria**
- [ ] Contact the owner of Nightwave Plaza ( https://www.patreon.com/nightwaveplaza )
- [ ] A way to play and pause the music
- [ ] A way to control the volume of the music

PR #280, opened 2020-05-03
React

Zemanzo:
Since we aim to have a lot more UI, it would be a great step forwards if we could use a frontend framework. For that reason, I have implemented a React pipeline.
Complete details can be found in the TDD, but in short: We create a base page on the server and attach a root component. The server renders the page (creates the HTML markup) and sends it to the client. The client can then hydrate the components to re-attach the interactivity to the components.

It also requires some new steps in the project setup:
- Server code is now also built and bundled. To run the server, first run `yarn build:server` and then `yarn start`.
- For a possible easier workflow, I added a `launch.example.json` for vscode. Change the filename (or make a copy) to `launch.json. This will allow you to run the debugger in vscode for the server. Simply hit F5 and it will build & run the server.
I'll add this to the wiki when the PR is merged.

I had to go for some suboptimal solutions here and there though:
- We still have to manually add an entry point to the `webpack.client.config.js` (would be nice if this was taken care of on a per-route basis).
- For each route, we need to manually hydrate the server-side rendered React components. I looked for a solution that does this for every page, but I couldn't find a good way to have a common client script that points to the corresponding RootComponent for every route.
- The server renders HTML during runtime. It's a minor performance impact that could be solved by saving the rendered HTML to disk (minify while we're at it) and serving that as a static page. Downside would be that it complicates any dynamic content, such as having a user logged in and displaying it on the page.
  *NOTE: It also rendered the HTML during runtime before this PR, so compared to that there's no performance hit.*

**Additional findings**
Whilst further experimenting and researching this topic, I stumbled upon the problem of passing server side data to the client. With SSR, we render the HTML on the server, which has to be hydrated on the client side to attach the correct interactivity (such as event listeners). This unfortunately also requires the data that was used to render it on the server. That means we'll have to somehow also pass along the raw data from the server to the client, which is rather wasteful since the data had already been rendered. A good example can be found here: https://developers.google.com/web/updates/2019/02/rendering-on-the-web#rehydration-issues

**Related issues**
Makes the following issues easier:
#199 
#197 maybe
#185 
#158 
#155 
#148 
#91 
#90 
#89 
#88


.vscode/launch.example.json
```diff
@@ -0,0 +1,12 @@
+{
```
Qaomen, line 1: Really useful addition! Though I was wondering why we wouldn't remove the `.example` bit for `launch.json` and `tasks.json`. That said, I don't know how often people customise those things for themselves, just thinking this could remove a small setup step.

scripts/webpack.client.config.js
```diff
@@ -1,42 +1,64 @@
+const sharedConfig = require("./webpack.shared.config");
+
 const path = require("path");
 const CopyPlugin = require("copy-webpack-plugin");
 const TerserPlugin = require("terser-webpack-plugin");
-const isProduction = process.env.NODE_ENV === "production";
+const isProduction = sharedConfig.isProduction;
```
Qaomen: Considering this variable is only used once, we might as well use `sharedConfig.isProduction` directly instead.

src/server/database/users.js
```diff
@@ -389,6 +389,25 @@ module.exports = function(db, common) {
...
+				stat_rounds_entered
+			FROM
+				users
+			ORDER BY
+				stat_points_earned DESC
+			LIMIT ?`
+		),
+
+		getTopAlltime(limit) {
+			return this._getTopAlltime.all(limit).map((entry, index) => {
```
Qaomen, line 411: I'd recommend filtering out entries with 0 points, since those users never participated in a race.

src/server/router/page.js
```diff
@@ -0,0 +1,109 @@
...
+			this._renderPage(req, res);
+		});
+	}
+
+	_renderPage(req, res) {
+		res.setHeader("Content-Type", "text/html; charset=utf-8");
+		res.setHeader("Transfer-Encoding", "chunked");
+
+		// Get latest props
+		this.props.serverSideProps = this._deepRunObjectFunctions(this.props.serverSideProps);
```
Qaomen: This overrides the original object, meaning any functions in them are replaced by the results, causing it to no longer update after it's called for the first time.

Actually, it may be better to create a class for a specific page that derives off of `Page`, and do this props/update function stuff in an overridable function, since that's what it ultimately tries to do here.
E.g. a `getLatestProps` function that, in the case of a `LeaderboardsPage` class, calls `db.user.getTopAllTime` and does the stuff it needs to do. Might be nice to cache the results and only call the database function(s) if a race has happened since the last time the page got requested.

If we want to do more stuff with caching, we could let `getLatestProps` return `true` when we want to rerender `rootComponentStream`, and otherwise save the last result and reuse it. (...But maybe I'm going on too much about caching at this point.)

src/server/router/page.js
```diff
@@ -0,0 +1,109 @@
...
+
+		// Create stream data
+		const htmlStart = `
```
Qaomen: Could/should we cache part of this in the constructor? The only thing that's not constant is what we put in `window.__INITIAL_STATE__`, but other than that it would save some string parsing.

Perhaps we could set `window.__INITIAL_STATE__` in `htmlEnd`? Looking at that article you linked (which mentions streaming) and how the client only hydrates when the DOM is ready, this could technically make for a faster FCP, I think?
(That said I'm only theorizing and have no actual idea so I'd love some input)

src/client/leaderboards/root-component.js
```diff
@@ -0,0 +1,26 @@
...
+import Leaderboard from "../components/leaderboard";
+
+const LeaderboardRoot = styled.div`
+	display: flex;
+	justify-content: center;
+`;
+
+const RootComponent = (props) => {
+	const serverSideProps = props.serverSideProps || window.__INITIAL_STATE__;
+	console.log(serverSideProps);
```
Qaomen: Leftover logging.

src/client/components/leaderboard.js
```diff
@@ -0,0 +1,51 @@
...
+`;
+
+const Leaderboard = (props) => {
+	return (
+		<LeaderboardWrapper>
+			<LeaderboardHeader>
+				{props.header}
+			</LeaderboardHeader>
+			<EntriesWrapper>
+				<LeaderboardEntry header rank="RANK" username="NAME" stat_points_earned="SCORE" stat_rounds_entered="ENTRIES" />
```
Qaomen, line 44: Considering the layout of the leaderboards and the way the header's style is currently handled, wouldn't using `table` (and `thead`) be a cleaner way of setting this up?

src/client/components/header.js
```diff
@@ -0,0 +1,56 @@
...
+
+const Header = (props) => {
+	return (
+		<header id="header">
+			<a href="/" id="title">
+				<img src="images/logo.svg" />
+				<h1>Manzo&apos;s Marbles</h1>
+			</a>
+			<nav id="internalLinks">
+				<HeaderLink
```
Qaomen, line 25: You're probably aware but just so we don't forget, there's no `imageLink` set yet so it's showing up blank.

src/client/contact/root-component.js
```diff
@@ -0,0 +1,22 @@
...
+const email = typeof config.contact.decode === "function"
+	? config.contact.decode(config.contact.email)
+	: config.contact.email;
+
+const RootComponent = () => (
+	<React.Fragment>
+		<img src="images/logo_text.svg" alt="Manzo's Marbles" id="headerImage" />
+		<section className="column">
+			<h2>Contact</h2>
+			<p>
```
Qaomen: React's validation on the client throws a warning because a `p` shouldn't have block-level elements in them. Changing this to a `div` seems to fix it without further issues.

src/server/router/page.js
```diff
@@ -0,0 +1,109 @@
...
+import headerProps from "./componentProps/header";
+
+export default class Page {
+	constructor(app, details, rootComponent, props = {}) {
+		this.details = details;
+		this.rootComponent = rootComponent;
+
+		// Add default props
+		this.props = props;
+		this.props.serverSideProps = this.props.serverSideProps || {};
```
Qaomen: Why use `serverSideProps` specifically? It looks like that's the only property in `this.props`.

src/web-marbles.js
```diff
@@ -158,9 +158,39 @@ app.get("/privacy", function(req, res) {
...
+	{
+		id: "leaderboards",
+		label: "Leaderboards",
+		description: "An overview of the current rankings for web-marbles"
+	},
+	LeaderboardsComponent,
+	{
+		serverSideProps: {
+			leaderboards: {
+				alltime: getLatestLeaderboard.bind(db) // oofies
```
Qaomen: ~~Big yikes.~~

src/server/router/page.js
```diff
@@ -0,0 +1,109 @@
...
+	_renderPage(req, res) {
+		res.setHeader("Content-Type", "text/html; charset=utf-8");
+		res.setHeader("Transfer-Encoding", "chunked");
+
+		// Get latest props
+		this.props.serverSideProps = this._deepRunObjectFunctions(this.props.serverSideProps);
+
+		// Create stream data
+		const htmlStart = `
+<!DOCTYPE html>
```
Qaomen, line 23: `htmlStart` (and I think `htmlEnd` too?) causes a warning during hydration (`Did not expect server HTML to contain the text node "    " in <div>`) due to the white-space it sends over. We should probably regex that white-space away so we at least keep the nice formatting in code.

src/client/components/leaderboard.js
```diff
@@ -0,0 +1,51 @@
+import React from "react";
```
Qaomen, line 1: I wonder if we should put these components in the related page folder instead (e.g. this file in `client/leaderboards` instead of `client/components`), since we'd only use it there.

scripts/webpack.client.config.js
```diff
@@ -1,42 +1,64 @@
...
 		]
 	},
 	plugins: [
+		...sharedConfig.plugins,
 		new CopyPlugin([
 			{
-				from: path.resolve(__dirname, "node_modules/three/examples/js/libs/draco/gltf"),
+				from: path.resolve(
+					__dirname,
+					"node_modules/three/examples/js/libs/draco/gltf"
```
Qaomen: The path should have `../` in front.

Qaomen (changes requested):
Finally got through it, sorry for the wait! Some comments may be more about me not completely understanding how React works or the reasoning why some things are set up the way they are, so bear with me there.

From what I've learned about React and how it's implemented here, I think I'm fine with the way it's set up. The main thing I'm unsure about is how the `Page` class is implemented, so that'd be worth discussing. Overall pretty hyped to see this feature getting done!

.vscode/launch.example.json
Zemanzo: I actually ran into this problem at work, where they are committed, hence I opted to use `.example` here.

What we could maybe do is create a very simple `yarn setup` script, that will copy all example (config) files, and ask questions where required to fill them in. That'd make setup a lot more straightforward, but would also create more stuff to maintain. Just a thought though.

src/client/components/leaderboard.js
Zemanzo: Would we though? I imagined maybe having a small variant in the client page (as one of the tabs in the left sidebar people could switch to, to quickly check) and maybe have a variant on the front page as well.

src/server/router/page.js
Zemanzo: Yeah that's a fair point, quite sure this should be beneficial and easy to do.

src/server/router/page.js
Zemanzo: I very much like this idea. Could you perhaps try to create this and commit it to this branch?

src/server/router/page.js
Zemanzo: Good point, on the RootComponent, this will always be the only property. When I wrote this, I still imagined we could have other props on it as well. I'll probably change this to simply use props then.

Zemanzo:
Should implement styled components properly for SSR https://styled-components.com/docs/advanced#server-side-rendering

src/server/router/page.js
```diff
@@ -0,0 +1,110 @@
...
+		this.details = details;
+		this.props = {}; // Should be filled by the page's _getLatestProps function
+
+		// Set up route in express
+		app.get(`/${details.id}`, (req, res) => {
+			this._renderPage(req, res);
+		});
+
+		// Add route to header props
+		headerProps.route = details.id;
```
Qaomen: This modifies the `headerProps` module, meaning all pages are affected by the page that changes this last.

src/server/router/page.js
```diff
@@ -0,0 +1,110 @@
...
+	_getLatestProps() {
+		return false;
+	}
+
+	/**
+	 * Minifies the HTML string, and returns the minified version
+	 * @param {string} html
+	 * @returns {string} Minified HTML string
+	 */
+	_basicMinifyHTML(html) {
```
Qaomen, line 107: Currently unused, but the whitespace warning is gone too so... I guess we don't need it?

Qaomen (changes requested):
Just some small things left!

src/server/router/page.js
Zemanzo: We do though! Think of all the wasted bandwidth! All that whitespace and commentary that people don't need... Just gotta wrap `this.htmlStart` with it.

src/server/router/page.js
Zemanzo: oh oopsie poopsie, good catch

Qaomen (approved):

PR #288, opened 2020-11-11
Discord domain update, and a dependency

Qaomen:
Discord has changed its main domain from `discordapp.com` to `discord.com` (excluding `cdn.discordapp.com`). The old domain still redirects nicely, but might as well update this!
A small change is included to fix a potential issue where member data (e.g. roles) aren't fetched yet when a user uses the client chat, skipping Discord's message event that would normally update the cache for that member.

Also wanted to update some deps because why not, but a few involve major changes so maybe some time later! :D


Zemanzo (approved):
Ship it!

PR #289, opened 2020-11-14
Three.js dependency update, rendering fixes

Qaomen:
**Changes**
- Set up a basic post-processing pipeline, added a gamma correction shader.
- Removed all gamma encoding in the texture objects.
- Added pre/post-render steps for the `Sky` object to apply the exposure settings.
- `renderCore` now has a main camera, which is shared between the camera types.
- Fixed the tracking camera rotating over the z axis.
- Fixed an issue where some world objects would cast shadows regardless of configuration settings.

**Referencing issues**
None :(

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
     "semver": "^7.1.0",
     "stats-js": "^1.0.0",
     "styled-components": "^5.1.0",
-    "three": "^0.114.0",
+    "three": "^0.122.0",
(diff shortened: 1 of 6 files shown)


Zemanzo (changes requested):
As mentioned, the way stuff looks in this PR is too different from how it used to be. So please change it to the way it looked before! (Or get real close, that's fine too.) (Or improve it! :D)

Zemanzo (approved):

PR #295, opened 2021-01-08
Race logic refactor

Qaomen:
Maybe this is more a big patch than a feature, but it's a primer for marble bots and (eventually) support for different game modes!
~~And for worker threads but let's not talk about that.~~

**Major changes:**
- All race-related logic has been moved from `game.js` to the `DefaultRace` class, along with some other changes:
	- Added the `STATE_PREPARING` game state, for any preparation logic needed right before the race start.
	- Marble attribute checks have been moved from `commands.js` to `DefaultRace`, and have been rewritten.
	- Added the `RaceEntry` class.
	- Database writes are now handled by race instances.
	- `game` can no longer directly spawn marbles, though `game.addRaceEntry` can still be used to spawn bots at any time.

**Minor changes:**
- Fixed colored logging not working on server-side worker threads.
- Fixed/cleaned up the physics module initialization (which also resolved a circular dependency).
- Physics world automatically starts/stops simulation based on any existing marbles.
- Added `physicsWorld.clearMarbles()`
- Other small changes / new helper functions to make the code less error-prone.


src/server/races/default-race.js
```diff
@@ -0,0 +1,279 @@
...
+	// Non-state functions
+	//
+
+	// Returns true if the marble/entry limit has been reached
+	addRaceEntry(id, name, attributeString = "") {
+		// Don't add new player entries if the limit has been reached, or players that already entered
+		// Exception: Added bots with a undefined id (only spawnable by devs/operators)
+		if(typeof id === "string" && this.isRaceFull()
+		|| (typeof id === "string" && typeof this.raceEntries.find(playerEntry => id === playerEntry.id) !== "undefined") )
+			return this.isRaceFull();
```
Zemanzo, line 118: As stated in our styled guide:
> For single line ifs, keep them in a single line `if(bool) myFunction();`.

In this case it's a pretty long line, so I would suggest just using brackets instead.

src/server/chat/commands.js
```diff
@@ -1,44 +1,17 @@
...
 			// Enter a marble
 			commandString: "marble",
 			permission: "ANYONE",
 			action: function(messageContent, id, username) {
-				game.addPlayerEntry(
+				game.addRaceEntry(
 					id,
 					username,
-					_createMarbleAttributesObject(messageContent)
+					messageContent.substr(7)
```
Zemanzo: *Technically* `substr()` is deprecated. Using `substring()` or `slice()` would be preferred.

Additionally, why 7? That's very confusing in this context. Why wouldn't the `addRaceEntry()` function take care of this? Why bother cutting down the string length?

If you're set on keeping the meaningless value, I would suggest putting it in a variable with a descriptive name such as `RACE_ENTRY_MESSAGE_LENGTH`. Preferably it would be dynamically set based on the commandString, or not be there at all.

src/server/chat/commands.js
```diff
@@ -85,10 +58,10 @@ const commands = function() {
 			action: function(messageContent) {
 				let amount = Math.min(100, parseInt(messageContent.substr(11)) || 10);
 				for (let i = 0; i < amount; i++) {
-					game.spawnMarble(
+					game.addRaceEntry(
 						undefined,
 						undefined,
-						_createMarbleAttributesObject(messageContent)
+						messageContent.substr(11)
```
Zemanzo: Same as above.

src/server/races/default-race.js
```diff
@@ -0,0 +1,279 @@
...
+		this.onNewRaceEntry(entry);
+
+		return this.isRaceFull();
+	}
+
+	// Returns a marbleAttributes object for the given user id and attributes string
+	parseMarbleAttributes(id, attributeString) {
+		let marbleAttributes = {};
+		let messageSections = attributeString.split(" ");
+		messageSections.length = Math.min(messageSections.length, 4); // At most, check only the first 4 words
```
Zemanzo, line 134: I don't mean to insult you but dear lord this is nasty. Can you not just do the `Math.min()` in the `for` loop declaration?

src/server/races/default-race.js
```diff
@@ -0,0 +1,279 @@
...
+
+		for(let i = 0; i < messageSections.length; i++) {
+			Object.keys(this.marbleAttributeList).forEach( attribute => {
```
Zemanzo: This generates a new array to loop over, inside of a loop. A better alternative would probably be to use a `for ... in` loop? e.g:
```js
  for (let attribute in this.marbleAttributeList) {
    // Do stuff...
  }
```

src/server/races/default-race.js
```diff
@@ -0,0 +1,279 @@
...
+
+	// Returns the current/expected marble count for the race
+	getProjectedMarbleCount() {
+		return this.raceEntries.length; // By default, we have one marble per entry
+	}
+
+	onMarbleFinished(entryId) {
+		let marble = this.marbleEntries[entryId];
+		if(marble.finished)
+			return;
```
Zemanzo, line 187: Use a single line, or brackets.

src/server/races/default-race.js
```diff
@@ -0,0 +1,279 @@
...
+
+		// Get raceEntry that belongs to this marble
+		let raceEntry = this.raceEntries.find((raceEntry) => {
+			return marble.userId === raceEntry.id;
+		});
+
+		if (raceEntry) {
+			// Add their best time to the entry (so it can be stored in the case of a PB)
+			if(typeof raceEntry.time === "undefined")
+				raceEntry.time = marble.time;
```
Zemanzo, line 209: Single line or brackets

src/server/races/default-race.js
```diff
@@ -0,0 +1,279 @@
...
+
+			// Award points if eligible
+			if(raceEntry.earnsPoints)
+				this.awardPoints(raceEntry, marble);
```
Zemanzo, line 214: Single line or brackets

src/server/races/default-race.js
```diff
@@ -0,0 +1,279 @@
...
+			return match === null ? undefined : match[0];
+		};
+		this.defaultMarbleAttributeList.color = function() {
+			return utility.randomHexColor();
+		};
+
+		this.marbleAttributeList.skinId = function(attribute, userId) {
+			let skinId = skins.idList[attribute];
+			if (typeof skinId === "undefined")
+				return undefined;
```
Zemanzo: Since the outcome of this condition is the same as the conditions below, it's better to combine them into a single `if`.

src/server/races/default-race.js
```diff
@@ -0,0 +1,279 @@
...
+		return this.isRaceFull();
+	}
+
+	// Returns a marbleAttributes object for the given user id and attributes string
+	parseMarbleAttributes(id, attributeString) {
+		let marbleAttributes = {};
+		let messageSections = attributeString.split(" ");
+		messageSections.length = Math.min(messageSections.length, 4); // At most, check only the first 4 words
+
+		for(let i = 0; i < messageSections.length; i++) {
```
Zemanzo, line 137: We can skip the first `messageSection` right? It will always be `!marble`.

src/server/races/default-race.js
```diff
@@ -0,0 +1,279 @@
...
+		this.pointContributors = 0;
+		this.marbleEntries = [];
+		this.marbleLimitReached = false;
+		this.marblesFinished = 0;
+
+		this.round = new Round(levelId);
+
+		// Contains a set of functions that returns a valid value if the passed attribute string can be parsed correctly
+		this.marbleAttributeList = {};
+		this.defaultMarbleAttributeList = {}; // Similar to above, but sets defaults for attributes that were not set/provided by the user
```
Zemanzo: Suggestion: Make a single list of marble attributes. Maintaining two lists containing the same value (but not entirely) is a bit messy to me. I'd be okay with it if they were generated based on a single list of objects, like this:
```js
const marbleAttributes = {
  color: {
    dynamicValue: () => {},
    defaultValue: () => {}
  },
  size: {,
    defaultValue: () => {}
  },
  // ...
};

for (let attribute in marbleAttributes) {
    this.marbleAttributeList[attribute] = marbleAttributes[attribute].dynamicValue;
    this.defaultMarbleAttributeList[attribute] = marbleAttributes[attribute].defaultValue;
}

```

src/server/races/default-race.js
```diff
@@ -0,0 +1,279 @@
...
+						marbleAttributes[attribute] = match;
+					}
+				}
+			});
+		}
+
+		// Set any default attributes that are (still) undefined
+		Object.keys(this.defaultMarbleAttributeList).forEach( attribute => {
+			if(typeof marbleAttributes[attribute] === "undefined") {
+				marbleAttributes[attribute] = this.defaultMarbleAttributeList[attribute]();
```
Zemanzo: There isn't really a need for the default attributes to be a function is there? I mean, I'm not opposed to doing it this way but it does feel a bit silly.

src/server/races/default-race.js
```diff
@@ -0,0 +1,279 @@
...
+					}
+				}
+				let record = gameConstants.RECORD_NONE;
+				if (personalBestIds.includes(player.id)) {
+					record = gameConstants.RECORD_PB;
+				}
+				scoreResults.push(player.id, player.pointsEarned, pointTotal, record);
+			}
+			if(scoreResults.length > 0)
+				this.game.onRaceResults(scoreResults);
```
Zemanzo, line 286: Single line or brackets

Zemanzo (changes requested):
Ohhh, this is getting me excited to make various game modes!

Aside from the inline comments, there's a lot of variables that can be a `const` instead of `let`. I didn't bother marking all of them, because I'd be busy for days. I'm okay if you leave them for now, but do at least take it into account for the future.

src/server/chat/commands.js
Qaomen: Didn't catch `substr()` being deprecated, so that will be changed. I will also change it for `!level`, since this isn't the first time this function is used in this file.

`substring()` returns the string starting from the given index (and optionally, and end index). So here, it removes `!marble` from the message since the command isn't part of the attributes (to add to that, `addRaceEntry` isn't aware of commands and requests only a string of attributes). I'll change it to base it off of the command string directly, and add a comment to clarify the purpose.

(I'm not sure what kind of assumption was being made here in the end, but if you believed this code was flawed to the point where it would undoubtedly break functionality, *please* confirm this by running the code next time. There are cleaner and more to the point ways to comment on a magic number and/or a unclear function call)

Qaomen: I'll also change the `substr()` calls for the rest of the codebase while I'm at it, since it's used more frequently than just here and well, it's deprecated.

src/server/chat/commands.js
Qaomen: See reply above. I'll also change it for the `for` loop logic that's part of this command.

src/server/races/default-race.js
Qaomen: Setting the `length` of a Javascript array directly looks a bit iffy, but it is a supported way to change the size of an array. If it helps you, I will change it to `splice(0,4)`, which does the exact same thing.

I decided to not do it in the loop here to keep the `for` loop itself cleaner while getting a bit more room to add a comment. We also don't call `Math.min` for every iteration in the loop that way, for what it's worth.

src/server/races/default-race.js
Qaomen: Is it better, or is it preference?
Personally, I don't like conditionals where a lot of different/unrelated evaluations are done at once, because it makes it harder to read and understand what is being checked (and in what order the checks need to happen, if applicable). Keeping this `if` as an early-out and the one below as purely a permission check seems like more readable code to me.

Also, combining these adds an `||` operator and requires the permission check to be wrapped in parentheses, making it even less readable in my opinion.

src/server/races/default-race.js
Qaomen: That was an old bug, which has been fixed by removing the marble/lotsofbots command using `substring()` in `commands.js`.

src/server/races/default-race.js
Qaomen: Good suggestion! This does make it easier to manage, so will be changed.

src/server/races/default-race.js
Qaomen: There already is a reason for this to be called as functions: the default randomized `color` and `size` which are defined above. Other game modes could also have reason to use this (e.g. for auto-joining teams, themed skins, whatever else it could come up with), so this adds flexibility for those cases.
I can understand the confusion with `default` part of the name though, so let me know if you have suggestions for alternatives.

src/server/races/default-race.js
Zemanzo: Fair points, we can keep it as it is 👍

src/server/races/default-race.js
Zemanzo: You're right, it looks very weird to me but it is a completely valid way of resizing the array. Feel free to leave it as is, I'll adapt to it! And fair point on not adding it in the `for` loop.

src/server/races/default-race.js
Zemanzo: Ah yes, oversight on my part. I expected the default to just be a static value. Not sure on any alternative names though 🤔

src/server/races/default-race.js
Qaomen: Normally this would be the better option, but in the current case, `attribute` is a `string`, which we use to write to the correct properties for the `marbleAttributes` object. The given alternative gives us an `Object` instead, which wouldn't make this possible.

src/client/generate-tiny-uuid.js
```diff
@@ -14,7 +14,7 @@ function generateTinyUUID() {
 	for (let i = 0; i < l; i++) {
 		let remain = decimal % charset.length;
 		decimal = (decimal - remain) / charset.length;
-		tUUID += charset.substr(remain, 1);
+		tUUID += charset.substring(remain, 1);
```
Zemanzo: `substring` is not a drop-in replacement here!

> _substring_ takes parameters as (from, to).
_substr_ takes parameters as (from, length).

The generateTinyUUID function no longer works correctly now. (Maybe we should use a proper package for this instead, but that's another discussion)

src/server/races/default-race.js
```diff
@@ -0,0 +1,291 @@
...
+				defaultValue: () => {
+					// 2% chance for a entry's marbles to be of a random size between .3 and .6
+					return Math.random() > .98 ? (.3 + Math.random() * .3) : 0.2;
+				}
+			},
+			skinId: {
+				userValue: (userId, attribute) => {
+					let skinId = skins.idList[attribute];
+					if (typeof skinId === "undefined")
+						return undefined;
```
Zemanzo, line 60: 👀 Single line, or brackets?

src/server/races/default-race.js
```diff
@@ -0,0 +1,291 @@
...
+
+		this.onNewRaceEntry(entry);
+
+		return this.isRaceFull();
+	}
+
+	// Returns a marbleAttributes object for the given user id and attributes string
+	parseMarbleAttributes(id, attributeString) {
+		let marbleAttributes = {};
+		let messageSections = attributeString.split(" ").filter( (attr) => {return attr != "";}); // Split, filter out additional whitespaces
```
Zemanzo: You can clean up the filter callback a bit:
`.filter(attr => attr != "")`

I'm also wondering about the non-strict check, can't we use `!==` in this situation?

src/server/races/default-race.js
Zemanzo: But it doesn't..? Unless I am missing something or being a big dumb (which is a very real possibility), a `for ... in` loop gives you the object keys as strings, not the values?

Zemanzo (changes requested):

src/server/races/default-race.js
Qaomen: Apparently I stand corrected! I'm kind of baffled that iterating over an object's properties this way gives you `string`s instead of, well, the properties. Js remains unintuitive and weird as hell for me, it seems.

Zemanzo: Just for complete clarity;
The `in` keyword gives you the keys of the object:
```js
for (let key in object) { ... }
```

~The `of` keyword gives you the values of the object~
That's not true actually, this only works for iterables such as arrays... In that case, use `Object.values()` or just use the `for (let key in obj)` and manually get the value such as `const value = obj[key]`.

Zemanzo (approved):

PR #298, opened 2021-01-12
Bugfix: Inactive tab / network buffer issue

### LINKED ISSUE #232: When client tab is hidden, the networking buffer increases (whilst it shouldn't)
**Expected behavior**
When the client tab becomes inactive, it should not adapt the network buffer because of the tab's low processing priority.

**Actual behavior**
When the client tab becomes inactive, it increases the network buffer because of the tab's low processing priority. This causes unnecessary delays.

**Reproduction steps**
1. Go to `/client`.
2. Start a marble race.
3. Go to a different tab in the same browser window for ~30 seconds.
4. Open the client tab again.
5. Network buffer has increased causing delays.

**Platform**
OS: Windows 10
Browser version: Firefox Developer Edition 68.0b11 (64-bit)
Build version: v1.2.0

**Related issues**
Might be related to #206

Qaomen:
Fixes the increasing network delay issue when the client page is an inactive tab for a while.

Previously, game/network updates were triggered by the rendering only. This generally worked fine, except that the rendering loop slows significantly when the tab isn't active, causing the network to still receive data while the update functions weren't called. This pull request adds a simple `updateManager` module to create a more centralized way of adding update functions to a list and allows any other module to trigger an update (instead of completely leaving it the renderer's responsibility).

Because the network packets stop right after the end of a race (as there's no new data to send at that time), there *is* a tiny bit of network buffer left at the end of a race that doesn't get processed immediately if the tab stays inactive. This would still cause a "delay", unless the client switches back to the tab again at some point or once the game state changes from `finished` to `waiting`. I'd consider this only a small issue, and given how the solution to that would involve creating a worker thread to periodically force updates I don't it's worth addressing.

**Referencing issues**
Closes #232

### DIFF
--- a/src/client/client/game.js
+++ b/src/client/client/game.js
@@ -5,6 +5,8 @@ import { userState } from "../user-state";
 import { renderCore } from "../render/render-core";
 import { cameras } from "../render/cameras";
 import { HUDNotification } from "./hud-notification";
+import { updateManager } from "../update-manager";
+import { networking } from "./networking";
 import * as gameConstants from "../../game-constants.json";
 
 let game = function() {
@@ -92,6 +94,11 @@ let game = function() {
 		}
 	};
 
+	let _update = function(deltaTime) {
+		levelManager.activeLevel.update(deltaTime);
+		networking.update(deltaTime);
+	};
+
 	return {
 		// Returns a Promise that resolves once initialization is complete
 		// Can be called multiple times but will initialize only once
@@ -110,6 +117,7 @@ let game = function() {
 					_DOMElements.raceLeaderboardAuthorName = _DOMElements.raceLeaderboard.getElementsByClassName("authorName")[0];
 					_DOMElements.resultsList = document.getElementById("resultsList");
 					_DOMElements.resultsListTemplate = document.getElementById("resultsListTemplate");
+					updateManager.addUpdateCallback(_update);
 				});
 			}
 			return _initPromise;
--- a/src/client/client/game.js
+++ b/src/client/client/game.js
@@ -9,8 +9,8 @@ import { updateManager } from "../update-manager";
 import { networking } from "./networking";
 import * as gameConstants from "../../game-constants.json";
 
-let game = function() {
-	let _audio = {
+const game = function() {
+	const _audio = {
 		start: new Audio("resources/audio/start.mp3"),
 		end: new Audio("resources/audio/end.mp3")
 	};
@@ -35,7 +35,7 @@ let game = function() {
 
 		_marbleBeingTracked = null;
 
-	let _trackMarble = function(marble, forceTracking) {
+	const _trackMarble = function(marble, forceTracking) {
 		if (forceTracking) {
 			renderCore.setCameraStyle(cameras.CAMERA_TRACKING);
 		}
@@ -61,7 +61,7 @@ let game = function() {
 	};
 
 	// Starts the "enter marbles now" visual timer. timeLeft in milliseconds
-	let _startEnterCountdown = function(timerValue) {
+	const _startEnterCountdown = function(timerValue) {
 		// Make sure it only runs once
 		if(_enterCountdownTimer !== null) clearInterval(_enterCountdownTimer);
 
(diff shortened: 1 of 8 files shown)


src/client/client/game.js
```diff
@@ -92,6 +94,11 @@ let game = function() {
 		}
 	};
 
+	let _update = function(deltaTime) {
```
Zemanzo: As discussed, we should start using `const` where ever we can. I'm not gonna push you to retroactively change it throughout the whole code base of course, but for anything new I will point it out and ask to change it.

src/client/editor/editor.js
```diff
@@ -112,6 +113,10 @@ EditorObject.prototype.setName = function(name) {
 let editor = function() {
 	let _activeTab = 2;
 
+	let _update = function(deltaTime) {
```
Zemanzo: Can be a `const`.

src/client/skins/skins.js
```diff
@@ -3,11 +3,12 @@ import { levelManager } from "../level-manager";
 import { marbleManager } from "../marble-manager";
 import { marbleSkins } from "../marble-skins";
 import { cameras } from "../render/cameras";
+import { updateManager } from "../update-manager";
 import domReady from "../dom-ready";
 
 // Set up core (rendering, level & marble management)
 renderCore.initialize(cameras.CAMERA_FREE);
-renderCore.updateCallback = function(deltaTime) {
+let update = function(deltaTime) {
```
Zemanzo: Can be a `const`.

src/client/update-manager.js
```diff
@@ -0,0 +1,29 @@
+import domReady from "./dom-ready";
+
+let updateManager = function() {
```
Zemanzo: Can be a `const`.

Zemanzo (changes requested):
As discussed before, we can manage the buffer better using the [page visibility API](https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API).

Aside from the variable declarations, this is pretty nice! Good to have these relatively small things fixed.

src/client/client/networking.js
```diff
@@ -6,17 +6,35 @@ import * as gameConstants from "../../game-constants.json";
...
+	let hiddenVarName, visibilityChangeEventName;
+	if(typeof document.hidden !== "undefined") {
+		hiddenVarName = "hidden";
+		visibilityChangeEventName = "visibilitychange";
+	} else if(typeof document.msHidden !== "undefined") {
+		hiddenVarName = "msHidden";
+		visibilityChangeEventName = "msvisibilitychange";
+	} else if(typeof document.webkitHidden !== "undefined") {
+		hiddenVarName = "webkitHidden";
+		visibilityChangeEventName = "webkitvisibilitychange";
```
Zemanzo: Prefixed variants aren't really necessary, see https://caniuse.com/?search=Page%20Visibility%20API

src/client/client/networking.js
```diff
@@ -6,17 +6,24 @@ import * as gameConstants from "../../game-constants.json";
...
 	let _timeDeltaRemainder = null; // Represents the timeDelta between two updates, or null if there are none
 	let _desiredBufferSize = config.defaultBufferSize; // Desired buffer size
 	let _previousMarblePositions = null;
 
-	let _processMessageEvent = function(event) {
+	// Page visibility logic. If the page is hidden, the network buffer will fast-forward when a race ends
+	// https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilityState
+	let isPageHidden = false;
+	document.addEventListener("visibilitychange", () => {isPageHidden = document.visibilityState === "hidden";}, false);
+	window.addEventListener("pagehide", () => {isPageHidden = document.visibilityState === "hidden";}, false); // Safari fix, which doesn't trigger the visibilitychange event
```
Zemanzo, line 24: These two callback functions are exactly the same. It might be nicer to store the function in a variable and use that twice, but I'll leave that up to you.

Zemanzo (approved):
Minor comment. You can change it if you'd like, but either way you can merge this PR. :)

Issue #267, opened 2019-12-06
Camera scrolling actions are jarring / non-functional on certain browsers


Zemanzo:
**Expected behavior**
Free camera: The moving speed of the camera should gradually change whilst using the mousewheel.
Tracking camera: The zoom level should change whilst using the mousewheel.

**Actual behavior**
Free camera: The moving speed of the camera changes very quickly, resulting in a loss of fine control.
Tracking camera: The zoom level does not change whatsoever.

**Reproduction steps for free camera**
1. Select the free camera setting.
2. Click and hold on the viewport.
3. Use WASD to move around, and start scrolling.
4. See jarring changes in movement speed.

**Reproduction steps for tracking camera**
1. Select the tracking camera setting.
2. Add a marble and start tracking it.
3. Scroll whilst having the cursor over the viewport.
4. Observe no zoom level changes.

**Platform**
OS: Windows 10
Browser version: Google Chrome v78.0.3904.108
Build version: 1.3.1

PR #302, opened 2021-09-18
Improved logic for zooming the tracking camera using the mousewheel

### LINKED ISSUE #300: Scrolling to zoom the camera in tracking mode does not work in some cases.
**Expected behavior**
When using the mousewheel in the viewport whilst the camera is in tracking mode, it should zoom in (or out) on the currently tracked marble.

**Actual behavior**
When using the mousewheel in the viewport whilst the camera is in tracking mode, it does not zoom. This is because the DeltaY value is way too big since it does not take the DeltaMode into account. Read more here: https://developer.mozilla.org/en-US/docs/Web/API/WheelEvent/deltaY

**Reproduction steps**
1. Just scroll the damn thing I basically explained it above already.

**Platform**
OS: Windows 10
Browser version: Firefox Developer Edition (93.0b5 (64-bit))
Build version: 1.3.1 (latest /dev branch)

Zemanzo:
Also added the ability to zoom with keyboard using the `+`, `-` and `0` keys (regular & numpad)

I've (successfully) tested this on the following platforms:
- Windows 10 - Firefox Developer Edition
- Windows 10 - Google Chrome
- OSX Big Sur - Firefox (using a trackpad)

**Referencing issues**
Closes #300 
Does not affect #267 because the free-camera is unaffected by this PR.

### DIFF
--- a/src/client/render/cameras.js
+++ b/src/client/render/cameras.js
@@ -347,6 +347,10 @@ function TrackingCamera(
 	const _YOffset = 7;
 	const _minZoom = .2;
 	const _maxZoom = 3;
+	const _zoomChangePerPixel = .002;
+	const _zoomChangePerLine = _zoomChangePerPixel * 16;
+	const _zoomChangeFineGrainModifier = .25;
+	const _zoomChangePerKeyboardStep = .2;
 
 	/**
 	 * Enables the controls
@@ -365,8 +369,8 @@ function TrackingCamera(
 				_self.distanceMultiplier = Math.min(_maxZoom, Math.max(_minZoom,
 					_self.distanceMultiplier + (
 						event.deltaY
-						* (event.deltaMode !== 0 ? 0.032 : .002)
-						* (event.shiftKey === true ? .25 : 1)
+						* (event.deltaMode !== 0 ? _zoomChangePerLine : _zoomChangePerPixel)
+						* (event.shiftKey === true ? _zoomChangeFineGrainModifier : 1)
 					)
 				));
 			}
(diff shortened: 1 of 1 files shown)


src/client/render/cameras.js
```diff
@@ -343,31 +343,58 @@ function TrackingCamera(
...
-			let newMultiplier = _self.distanceMultiplier + .05 * event.deltaY;
-			if (_self.enabled === true && newMultiplier > _minZoom && newMultiplier < _maxZoom) {
-				_self.distanceMultiplier = newMultiplier;
+		// 1 pixel should result in a .002 zoom change. When `deltaMode` is not 0, we assume notched intervals, of 16px each.
+		const zoomWheelFunction = function(event) {
+			if (_self.enabled === true) {
+				_self.distanceMultiplier = Math.min(_maxZoom, Math.max(_minZoom,
+					_self.distanceMultiplier + (
+						event.deltaY
+						* (event.deltaMode !== 0 ? 0.032 : .002)
```
Qaomen: I'm not really fond of these magic numbers. Could we add these to the `const`s above? Just like the camera movement speed, I'd imagine we may want to make this configurable at some point.

src/client/render/cameras.js
```diff
@@ -343,31 +343,58 @@ function TrackingCamera(
...
+		const zoomKeypressFunction = function(event) {
+			if (_self.enabled === true) {
+				switch(event.code) {
+				case "Equal":
+				case "NumpadAdd":
+					_self.distanceMultiplier = Math.max(_minZoom, _self.distanceMultiplier - .2);
+					break;
+				case "Minus":
+				case "NumpadSubtract":
+					_self.distanceMultiplier = Math.min(_maxZoom, _self.distanceMultiplier + .2);
```
Qaomen: Magic numbers, same as above.

Qaomen (changes requested):
I had no idea it was broken, but glad you fixed it! :D The added key events are pretty neat too.
Just a minor gripe about the hard-coded numbers, but other than that it's all good.

Qaomen (approved):
Now I feel bad for making the wizard's code less magical. :(
But it's good to go now! :D

PR #303, opened 2021-09-18
Update dependencies

Zemanzo:
More dependency updates! Woo! Hopefully this also fixes any open security issues when it comes to dependencies.

Only `Cookies` had breaking changes that affected us, so that logic has been updated.

`Discord.js` is not updated in this PR because that's done in #301 of course.


src/client/user-state.js
```diff
@@ -1,19 +1,19 @@
...
-		userState.data = Cookies.getJSON("user_data");
+		userState.data = JSON.parse(Cookies.get("user_data"));
 	}
 }, false);
 
 let userState = {
 	AUTH_CHANGED: 0,
 
-	data: Cookies.getJSON("user_data")
+	data: JSON.parse(Cookies.get("user_data"))
```
Qaomen: image
When I logged out on the client page and refreshed, the page failed to load fully and got this in my console.

package.json
```diff
@@ -44,16 +44,16 @@
...
+    "react": "^17.0.2",
+    "react-dom": "^17.0.2",
     "reconnecting-websocket": "^4.4.0",
     "request": "^2.88.0",
     "request-promise-native": "^1.0.7",
     "semver": "^7.1.0",
     "stats-js": "^1.0.0",
     "styled-components": "^5.1.0",
-    "three": "^0.123.0",
+    "three": "^0.132.0",
```
Qaomen, line 56: It looks like Three.js has breaking changes (which we get warnings about when building the client), causing the editor to break:
image

Qaomen (changes requested):
Wooh more dependency updates! Slightly less wooh is that the cookies and editor are borked, meaning I couldn't fully test the pako dependency update either. :(

Zemanzo:
Heck I'm so smart, probably didn't even rebuild the client after bumping version numbers but before testing.

Zemanzo:
@Qaomen Actually before you approve this, try this map:
[Beach2-alpha.zip](https://github.com/Zemanzo/web-marbles/files/7230464/Beach2-alpha.zip)

When you load it, the start & starting gate are slightly offset, until you click the world tab, which causes it to snap back to how they *should* be positioned.
Unfortunately, when playing the map, it seems the start is actually saved with the offset as marbles fall off the track when they are spawned. I had trouble with getting the `_combineGeometry()` function in **models.js** working nicely again whilst fixing dependency stuff. Maybe you could take a quick look and see what's wrong?

